PR #3, opened 2015-12-01
Update README.md

dol-leodagan:
Update Appveyor Badge
New CI Project


dol-leodagan:
This is all valid for continuous integration on DOL project with Versioning similar to Subversion Revision.

Nuget restore is now needed to build project, documentation should be updated to match this behavior

PR #4, opened 2015-12-03
Nunit Test restored

dol-leodagan:
Restore Unit Test with Nunit Framework.

Allowing Continuous Integration with Server Startup Test.


Graveen:
Ah yes i remember those good ole UTs. We never used them as a quality tool (DOL was not test-driven), but for mainly for debug. Thank you for the fix.

dol-leodagan:
Well at least it can offer some basic automatic validation tasks, especially with some features like Runtime Compilation of Game Server Scripts which was one of our pitfall on legacy code updates !

PR #8, opened 2016-03-08
Changed PacketProcessor (and Preprocessor) to have a non-static imple…

ElliotB256:
…mentation. This opens the way to having client specific Client->Server handlers. To speed up the implementation caching is used to save IPacketHandler lists for different versions - assemblies are only searched for the first login, thereafter the cache is used.

This would be a nice feature to work towards. For example, rather than having a single PlayerPositionUpdateHandler that has hard-coded logic to switch different client versions, we could have a separate v168.PlayerPositionUpdateHandler, v192.PlayerPositionUpdateHandler etc.

### DIFF
--- a/GameServer/packets/Client/168/PlayerHeadingUpdateHandler.cs
+++ b/GameServer/packets/Client/168/PlayerHeadingUpdateHandler.cs
@@ -55,7 +55,10 @@ namespace DOL.GS.PacketHandler.Client.v168
 			client.Player.TargetInView = ((flags & 0x10) != 0);
 
 			byte[] con = packet.ToArray();
-			if (!client.Player.IsAlive)
+            con[0] = (byte)(client.SessionID >> 8);
+            con[1] = (byte)(client.SessionID & 0xff);
+
+            if (!client.Player.IsAlive)
 			{
 				con[9] = 5; // set dead state
 			}
--- a/GameServer/packets/Client/168/PlayerPositionUpdateHandler.cs
+++ b/GameServer/packets/Client/168/PlayerPositionUpdateHandler.cs
@@ -620,8 +620,11 @@ namespace DOL.GS.PacketHandler.Client.v168
 			//**************//
 
 			byte[] con168 = packet.ToArray();
-			//Riding is set here!
-			if (client.Player.Steed != null && client.Player.Steed.ObjectState == GameObject.eObjectState.Active)
+            con168[0] = (byte)(client.SessionID >> 8);
+            con168[1] = (byte)(client.SessionID & 0xff);
+
+            //Riding is set here!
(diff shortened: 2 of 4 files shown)


ElliotB256:
I've also added the previously discussed Heading and Position fixes

Graveen:
Thank you !

@dol-leodagan it seems Travis is not accepting it? i'm merging btw, since the error is related to UnitTest / Lang (and since AppVeyor is building correctly for Windows).

dol-leodagan:
Yeah my PR #9 was fixing the Travis Build, I would have like to re-run the checks on this PR to make sure it was fixed...

ElliotB256:
Err, I sort of assumed there would be a bit more of a discussion phase before the accept! I wanted to say that I've tested this for two client versions (sorry, no more available to me!) for multiple log ins and they seem to work fine on my version. Please let me know asap if these changes cause any problems!

ElliotB256:
I agree with Leo's statement, is there a way to re-run the checks?

dol-leodagan:
They were run on the Merged Repository : https://travis-ci.org/Dawn-of-Light/DOLSharp

So at least compilation test and server startup is checked, but we don't have Unit Test for Packets.

You can open a new Pull Request if you intend to continue the work even on the same branch, if we keep the PR un-merged every commit on your branch will trigger new builds on CI.

ElliotB256:
I'm happy to do whatever you think is easiest. I intend to work on this again at some stage but probably sporadically/as time permits. Just let me know what you decide.

Re: Unit testing, did we have them before for packets?

Graveen:
Nope. We can't even said we got some UT in the DOL code, it was more a POC of wannabe UT, that was instanciating some classes and tryed an assert.

ElliotB256:
Has this change caused any problems so far?

ElliotB256:
I'll go ask

NetDwarf:
I get this error message with client version 1.109 and 1.117: 

"Preprocessor prevents handling a packet with packet.ID=x"
where x is 29,66,67,172,220 so far (http://imgur.com/2ky4m5B)

Most of the time I can log in, but sometimes I get kicked on log in.

Have no visibility issues, but I don't exactly know with which version I could trigger it.

ElliotB256:
Hi NetDwarf,

The client sends a bunch of updates such as player position update, etc, before the server has added the player to world, ie before these packets can be handled in a valid way. These were always rejected beforehand, but silently without any logging - hence I added a log message to make these visible. They are not neccessarily anything to be alarmed about :). To improve the message I could add the enum string representation for Client->Server packet id to the log message.

ElliotB256:
22 = 0x16
66 = 0x42
67 = 0x43
172 = 0xAC
220 = 0xDC

None of the above are in eClientPackets but these packet codes were received from the client - some unknown packets?

ElliotB256:
Lastly, is kick on login a new bug?

PR #11, opened 2016-03-28
* By Elcotek - Falcon Eye Ability handling for new Archery

elcotek:
After several tests of Falcon Eye with new Archery on my Server, Falocon Eye dont work with new Archery.

I hope you like it:)

### DIFF
--- a/GameServer/serverproperty/ServerProperties.cs
+++ b/GameServer/serverproperty/ServerProperties.cs
@@ -1732,6 +1732,11 @@ namespace DOL.GS.ServerProperties
 		[ServerProperty("classes", "allow_old_archery", "Should we allow archers to be able to use arrows from their quiver?", false)]
 		public static bool ALLOW_OLD_ARCHERY;
 
+        /// <summary>
+        /// Bow Base Crit chance for new Archery
+        /// </summary>
+        [ServerProperty("rates", "bow_base_critchance", "Bows Crit base percent. ", 10)]
+        public static int BOW_BASE_CRITCHANCE;
 
 		#endregion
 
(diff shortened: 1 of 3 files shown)


Graveen:
Thank you.we should base the ability on classes ID directly in the code, but also let the realm abilities 
Also, no need the *by Elcotek, you are automatically credited through your GH account name :)

Thank you for the commit.

dol-leodagan:
And for me it's a big **Warning**:

Don't add anymore RA specific code in SpellHandler.cs !

```csharp
((GamePlayer)m_caster).GetAbility<FalconsEyeAbility>()
```

We have too much of these static checks already we should not bloat code in area where it's already hard to maintain...

Try to override the methods from SpellHandler in Archery or ArrowSpellHandler, or try to bind to an Event Handler.... (Bind to GameLivingEvent Attack/Cast when using FalconsEye and change the critical formula per-swing ?) 

Do not use reference to Sub-class in "Super Class"

```csharp
if (this is ArrowSpellHandler)
```

This mean we can't compile, or re-use SpellHandler without having ArrowSpellHandler in the code base...

Is there a reason this only work with players and not with living ?
We should try to make all Skills to work with any Living (except if you need some specific properties  to a Player Object, like a DB Object or a Network Object...)

```csharp
if (m_caster is GamePlayer)
```
----
I know there are a lot of area where code is already set like this but we must start to raise DOL quality or we won't be able to have proper Unit Tests and other effective Modular programming...

elcotek:
I will do, i have now remove the last changes and will work on a new way 
to implemet.:)

thank you.

Am 29.03.2016 um 07:44 schrieb dol-leodagan:

PR #18, opened 2016-05-29
- Bugfix: Dragon should now properly reset AttackState when all enemies are gone.

NetDwarf:


Graveen:
Thank you for the fix

PR #26, opened 2016-06-24
Game database access

dol-leodagan:
Changing most game Database Access to use the new API using Queries with Parameters
Update some Save/Delete/Add Method to use collection instead of foreach


dol-leodagan:
Tests went OK on Storm Clone.

PR #34, opened 2016-12-31
Create OnlineStatusUpdate.cs

Shauzia:
This script is a modified serverlistupdate for website player count pinger. Simply edit user, pass and url then place 3 php files in your ftp directory.

### DIFF
--- a/GameServerScripts/startup/OnlineStatusUpdate.cs
+++ b/GameServerScripts/startup/OnlineStatusUpdate.cs
@@ -1,0 +1,245 @@
+/* 
+ * This script is for pinging your player count to a website. Take the time to edit some of the fields
+ * 
+ * Author: Dawn of Light
+ * Orginal Date: 31th December 2016
+ * Released by: Shauzia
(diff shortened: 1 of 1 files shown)


Graveen:
Thank you Shauzia

hunabku94:
In my oppinion, the place of this file is in User Files.

Graveen:
Or in GameServerScripts. I prefer it because user files are greats, but they finally end up in non-visible / non-maintained scripts.

Graveen:
Ok this is suitable for GameServerScripts alongside with WebUIupdate.cs - and i like the WebClient example available.
Shauzie (or anyone) can somebody include a serverinfo.php example ? Thus, we have both client AND server side example.

Shauzia:
Here's the php files - Made a little tutorial to go with it
[phpfiles.zip](https://github.com/Dawn-of-Light/DOLSharp/files/819458/phpfiles.zip)

Graveen:
Ok merged. I'll upload the files in the next commit. Thank you

PR #39, opened 2017-03-02
Duplicate claimedkeep entry fix

OntheDOL:
Fixes a bug introduced from a previous commit. Currently, when claiming
a keep the guild keep count is added twice. Removing this line allows a
guild to claim and release keeps, and claim again without server reboot.


dol-leodagan:
I'll review this soon, sorry for the delay...

dol-leodagan:
Nice fix !

PR #41, opened 2017-03-03
Added check for players attacking mobs while immune

OntheDOL:
Currently, a player who has pvp immunity can still attack mobs, but mob
cant fight back until immune is over. This adds a check and sends player
the denied message. More of an issue for pvp server mode where immunity
timers can be long.


Graveen:
Ok, merged. 
Please consider converting the "You can't attack..." sentence as a constant in Language Manager (this file: https://github.com/Dawn-of-Light/DOLSharp/blob/master/GameServer/language/EN/GameObjects/GameNPC.txt ).
Thank you :)

PR #42, opened 2017-03-04
Adds working resurrection sickness

OntheDOL:
Adds RvR, PvP, and PvE types of resurrection. RvR sickness is not curable except by Perfect recovery RA, and perfecter ML8 ability. PvP sickness player looses 3 con on death, curable. Con loss can be removed in server properties. PvE sickness removes con depending on number of death that level, curable.

### DIFF
--- a/GameServer/GlobalConstants.cs
+++ b/GameServer/GlobalConstants.cs
@@ -2458,10 +2458,10 @@ namespace DOL.GS
 		
 	}
 
-	public class GlobalSpells
+	public static class GlobalSpells
 	{
-		public const string PvERessurectionIllnessSpellType = "PveResurrectionIllness";
-		private static Spell m_PvERezIllness = null;
+		public const string PvEResurrectionIllnessSpellType = "PveResurrectionIllness";
+		private static Spell m_PvERezIllness;
 		public static Spell PvERezIllness
 		{
 			get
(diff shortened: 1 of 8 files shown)


OntheDOL:
Ok Disclaimer!
I have no coding experience before DOL - I'm a noob. So i am sure that this has flaws. :)
Anyway, DOL had some existing code for Res sickness, but it didnt work. Spellhandler would be found but no effect would apply to player ( Regardless of what is in DB) and no RvR sickness.

I have pretty much piggybacked off the existing code in various files as you can see. Added separate RvR sickness handler because this spell cannot be cured by healer and needs a different spell reference. Added PvP res sick which takes 3 con from player on death as well(Live Like). But added option in server properties to remove this for people who don't want. PvE res sick works as pve res sick should

OntheDOL:
I have tested this on PvP server settings dying to mob, player, releasing, rezzing etc, works fine
tested in Normal server settings dying to mob, player, releasing rezzing etc, works fine

OntheDOL:
Issues i know of - i used player.CastSpell instead of ISpellHandler StartSpell. I couldnt get this way to work. But there is nothing displayed about the spell casting in player chat box so maybe its fine.
- If player logs out while res sick, if log back in res sick is gone. I dont know how to fix this.
- Perfecter ML8 cure res sick. I dont know how to get ML abilities lol so i couldnt test this

Graveen:
We are glad you are giving an hand ! The code style need to be decent, but as you are mainly bugfixing, the entry ticket is affordable, so don't bother. I'd had more criticism on some hot parts such DB handler, global spell handlers and so on :)

Graveen:
And by the way, the patch you provide is elegantly coded, commented. Totally suitable. You have mastered the main C# tools (syntax, enums,...) and the DOL ones (langmgr, serverproperties...). Kudos !

PR #47, opened 2017-04-14
Bugfix: Champion Level spell cast speed and effectiveness

Xanxicar:
Fixed Champion Level spells cast speed.
Fixed Champion Level spells effectiveness:
CL specialization level will be returned as 50 again.
CL spell effective level should be calculated correctly again.

Due to changes to the naming of Champion lines the value provided by the global constant would never match. The constant's name has been changed, it's value set to match current CL line naming and an additional string.StartsWith() check has replaced a dysfunctional comparison.


Graveen:
Seems totally legit, thank you !

PR #56, opened 2017-05-21
Test PR

Graveen:
Dummy change


dol-leodagan:
Graveen you shouldn't sync your own fork with pull requests...

Just keep your master branch sanctuarized to rebase by using a local merge against upstream, and create a branch as soon as you want to edit any file.

DOLServer:
Yes, this commit was better (there is exactly in it what i want to be a PR) but i don't understand why i have the changelog for 22 june 2016. I'm using Tortoise as a GUI client - mainly because of the file mark + easy diff + easy targetted revert - so i would complete all the operation with a single tool, but this is not working as i expect.  I have created others accounts for a more tests, this is mandatory :D

PR #62, opened 2017-06-26
Grammar Correction in AbstractTask

jaystewart1976:
"I have no tasks for you at the moment. Come back sometime later, perhaps then you can help **we** with something."

Should read 

"I have no tasks for you at the moment. Come back sometime later, perhaps then you can help **me** with something."


Graveen:
Thank you Jay !

PR #63, opened 2017-06-27
Spelling correction in WolfPeltCloak.cs

jaystewart1976:
Should be Seamstress Lynnet not Lynnie

Evidence [http://camelot.allakhazam.com/quests.html?cquest=3](url)


Graveen:
TY

PR #64, opened 2017-07-29
Serverlistupdate from DB issue

jaystewart1976:
`Enable in-built serverlistupdate script FALSE`
With serverlist_update set to False, the console still shows the below.

`[16:09:48,656] Login/URL Error: Validating Arguments Failed! The server's player count will not be added/updated on the list!
[16:09:48,660] Login/URL Error: The arguments that failed validation are:
[16:09:48,661] Username
[16:09:48,662] Password`

This is using the latest SVN with nothing custom. 
So I added a check.

### DIFF
--- a/GameServerScripts/web/OnlineStatusUpdate.cs
+++ b/GameServerScripts/web/OnlineStatusUpdate.cs
@@ -95,8 +95,11 @@ namespace DOL.GS.GameEvents
 		[ScriptLoadedEvent]
 		public static void OnScriptCompiled(DOLEvent e, object sender, EventArgs args)
 		{
-			Init();
-		}
+		    if (ServerProperties.Properties.USE_SERVER_LIST_UPDATE_INTEGRATED.Equals(true))
+		    {
+		        Init();
+		    }
+        }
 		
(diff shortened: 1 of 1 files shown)


OntheDOL:
one thing with boolean values on csharp is you dont need to have equals(value)
simply having it as:
if (ServerProperties.Properties.USE_SERVER_LIST_UPDATE_INTEGRATED)
      dostuffifTrue();
is a "true" check
using an exclamation mark at the start of the boolean value is a "false" check
if (!ServerProperties.Properties.USE_SERVER_LIST_UPDATE_INTEGRATED)
    dostuffifFalse();

on a side note, if Graveen you read this, or if someone can contact Graveen, i think the email forwarding service on dolserver.net has been down for sometime. I cannot recover my forum account for new password or create new account , and have not received any emails from private messages for over a month now. Maybe you can look into that :)

jaystewart1976:
You can call me old fashioned!. The yellow spam annoyed me so I made a rapid change and apart from **readability**, They're functionally equivalent I think. Dont give him his account back for another month imo lol

PR #65, opened 2017-07-29
Serverlistupdate issue

jaystewart1976:
With serverlist_update set to False, the console still shows the below.

[16:09:48,656] Login/URL Error: Validating Arguments Failed! The server's player count will not be added/updated on the list! [16:09:48,660] Login/URL Error: The arguments that failed validation are: [16:09:48,661] Username [16:09:48,662] Password

This is using the latest SVN with nothing custom.


Graveen:
TY

PR #66, opened 2017-10-09
Fix Project to Compile for Mono/Linux

benfalk:
This corrects the case for several of the underlying projects so
the project can compile with the latest Docker image of `mono:latest`.
While a Windows system does not care about the case of a filepath Linux
does.


Graveen:
Thank you. 
Why especially theses files and not the others ? Theses are the ones in Game Server Scripts ?

benfalk:
These are all the offending lines in the project file that cannot be found with Mono's `msbuild` because the case does not match the source path. As you can see with the attached screen-shot a sample of the errors.

dol-mono-compile-errors

Looking at one example:

```
CSC : error CS2001: Source file '/opt/DOL/GameServer/gameobjects/SiegeWeapon/gameSiegeTrebuchet.cs' could not be found. [/opt/DOL/GameServer/GameServer.csproj]
```

The project file states that the files case is `gameSiegeTrebuchet.cs`; however the case is actually all lower in the source of `gamesiegetrebuchet.cs`.  Here is a link to the source as evidence:
https://github.com/Dawn-of-Light/DOLSharp/blob/master/GameServer/gameobjects/SiegeWeapon/gamesiegetrebuchet.cs

If you want I can revert the project files and instead correct the case of all the offending files; however, this seemed more simple and less noisy.

benfalk:
PS: I tried to register on the forum but never get a register email =\

Graveen:
Ok thank you. I see. For the forum PM me your pseudo i'll activate it manually, we have an issue with SourceForge.

dol-leodagan:
For me it's fine to change the project file instead of renaming a lot of files. (it will prevent breaking other pending PR)

We can always rename source files iteratively to CamelCase later...

Graveen:
TY

PR #68, opened 2017-11-03
Fixed Exception when adding first item

tegstewart:
Changed to create loot using an existing ItemTemplate from the DB.  Creating a new one causes the server to try to add the ItemTemplate to the ItemTemplate table, causing an exception which then mucks up the player's inventory.

Also changed LootGeneratorAurulite to be in the DOL.GS rather than DOL.GS.Scripts namespace.


tegstewart:
Oops.  Messed this up a little.  Closing and redoing.

PR #69, opened 2017-11-03
Fixed LootGenerator exceptions adding first item

tegstewart:
Changed to create loot using an existing ItemTemplate from the DB. Creating a new one causes the server to try to add the ItemTemplate to the ItemTemplate table, causing an exception which then mucks up the player's inventory.  This only happens if there isn't already an item of that type in the player's inventory, otherwise it just increases the count instead.

### DIFF
--- a/GameServer/gameutils/LootGeneratorAtlanteanGlass.cs
+++ b/GameServer/gameutils/LootGeneratorAtlanteanGlass.cs
@@ -17,10 +17,13 @@
  *
  */
 using System;
-
+using System.Reflection;
 using DOL.GS;
+using DOL.GS.PacketHandler;
 using DOL.AI.Brain;
 using DOL.Database;
+using DOL.Language;
+using log4net;
 
 namespace DOL.GS
 {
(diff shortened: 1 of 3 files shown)


Graveen:
What kind of exception did it throws ? I'm curious - it should not.
The proper way would be to handle the non existant id templates i guess, and do not fire the creation if not existant (so both your checks and the actual instanciation).
TY :)

tegstewart:
It's easier to describe how the process works.

I create a new dragon scale item template based on the one in the ItemTemplate table, and pass it into the loot list.  The DB code creates the item and puts it in the player's inventory, but then sees the new item template and tries to add it to the ItemTemplate table.  This violates unique primary key restrictions and throws an exception.  You end up with invisible items in inventory, and picking things up afterwards puts put the picked up in the slot the scales are supposed to be in.  When you relog, you get an error because you have multiple items in one slot, and while clearing the slot and relogging fixes it, it's REALLY annoying.

If the player already has a dragon scale in their inventory, the maxcount on it is incremented and it's fine.  It's only a problem when a new item is placed in a slot, and it's especially bad if you have multiple scales in the loot list.  If you have five scales, you'll end up with five separate invisible scales, each one in a different slot.

If I instead pull the ItemTemplate I want directly from the ItemTemplate table rather than making a new one, the DB code creates and adds the item, sees that the ItemTemplate doesn't need to be inserted, and it's all good.

I think it has to do with the dirty flag, but I have no idea how the DB code handles that stuff.  I just noticed that Mob.Drop() doesn't have this problem and was using the item template directly rather than making a new copy of it.

Graveen:
Ok i see. I'll wait and let it open for more comments to see if no side-effects.

PR #70, opened 2017-11-03
Correct Guild Form Behaviour

jaystewart1976:
Hi all, a rework on the /gc form.

> To form a guild, you must first form a full party of eight players, none of whom can already be guided. The person who is going to be the guild leader must visit the Guild Registrar NPC, found in the major cities of your realm. Each person must be in the same zone as the Guild Registrar. The group leader must select the Guild Registrar and type /gc form. Note that your guild's name must be unique. You cannot use a guild name that is already in use on your server. It costs one gold piece to start a guild. The money is deducted from the pocket of the character speaking to the Guild Registrar. Each member of the guild will receive a dialogue box asking if they accept membership in the new guild. Once all members have accepted, the guild is formed.

[official notes](http://darkageofcamelot.com/content/guild-guide-getting-started)

### DIFF
--- a/GameServer/commands/playercommands/guild.cs
+++ b/GameServer/commands/playercommands/guild.cs
@@ -49,11 +49,12 @@ namespace DOL.GS.Commands
         /// <returns></returns>
         public static bool IsValidGuildName(string guildName)
         {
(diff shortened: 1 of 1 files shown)


jaystewart1976:
if (!Regex.IsMatch(guildName, @"^[a-zA-Z àâäèéêëîïôœùûüÿçÀÂÄÈÉÊËÎÏÔŒÙÛÜŸÇ]+$") || guildName.Length < 0)
            {
                return false;
            }
            return true;

Not sure why its showing symbols that dont reflect the above ?

jaystewart1976:
Incorrect Encoding on guild.cs.

PR #72, opened 2017-11-04
Correct Guild Form Behaviour

Loki1976:
**guild.cs in current SVN is in ANSI encoding not UTF-8 it also doesn't use the tab spacing so this PR looks odd. I am not sure how to correct this issue with formatting sadly.**

Hi all, a rework on the /gc form.

To form a guild, you must first form a full party of eight players, none of whom can already be guided. The person who is going to be the guild leader must visit the Guild Registrar NPC, found in the major cities of your realm. Each person must be in the same zone as the Guild Registrar. The group leader must select the Guild Registrar and type /gc form. Note that your guild's name must be unique. You cannot use a guild name that is already in use on your server. It costs one gold piece to start a guild. The money is deducted from the pocket of the character speaking to the Guild Registrar. Each member of the guild will receive a dialogue box asking if they accept membership in the new guild. Once all members have accepted, the guild is formed.

[Official Notes](http://darkageofcamelot.com/content/guild-guide-getting-started)

### DIFF
--- a/GameServer/commands/playercommands/guild.cs
+++ b/GameServer/commands/playercommands/guild.cs
@@ -19,7 +19,7 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
-
+using System.Text.RegularExpressions;
 using DOL.Database;
 using DOL.GS.Keeps;
 using DOL.GS.PacketHandler;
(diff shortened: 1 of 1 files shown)


Graveen:
It's better. Thank you. Open a little for peer reviewing :)

OntheDOL:
did some tests with Jaystar with this a few weeks back, worked well!

Graveen:
excellent ! Thank you

PR #75, opened 2017-11-06
Remove Quit Timer

tegstewart:
I hate waiting to quit, so this change checks for ServerProperty "remove quit timer" and lets privlevel 1 accounts quit instantly so long as they aren't in combat.

### DIFF
--- a/GameServer/gameobjects/GamePlayer.cs
+++ b/GameServer/gameobjects/GamePlayer.cs
@@ -671,12 +671,7 @@ namespace DOL.GS
 			if (Client.Account.PrivLevel > 1) // GMs can always insta quit
 				bInstaQuit = true;
 			else if (Client.Player.InCombat == false)  // Players can only insta quit if they aren't in combat
-			{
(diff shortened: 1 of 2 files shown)


Graveen:
I hate waiting too ! :D
But it brings at least one very big issue:
* i'm roaming in NF, i saw a group incoming - faster than me - i can /q quickly enough and relog later
The /qtd was interesting: 
* you quit instantly
* you are still In game for 20s (in not in combat) / 1mn (if in combat)

So not sure about this PR. On the other side, i think i could accept because of the SP that is opening an highway for improvements w/o harming the actual way it works.

tegstewart:
The challenge is that quit times are hard coded, so a PR here is the only way to give the option for an instaquit to the community.

It's a useful enough feature that I think it should be available, but it definitely has the potential to cause problems so it can't be enabled by default.  I'm finding it incredibly helpful just from a testing standpoint.

Graveen:
Seems a fair trade. FALSE by default, then up to GMs to activate or not.

PR #76, opened 2017-11-06
Fixed dragon debuff damagetype

tegstewart:
Cuul and Gjal were using heat for their debuff damagetype, likely due to be copy/pasted from Gole.  Changed to match their glare and breath damage types.


tegstewart:
Travis is a necessary evil.  He's the supervillain that holds the world together and is a dick about it.

Graveen:
Thank you :)

PR #78, opened 2017-11-06
Fixed Hib ImportantDelivery

tegstewart:
There were problems with the Hib version of ImportantDelivery.  GameEventMgr.AddHandler() was throwing an exception when trying to add hooks when the DB had incorrect classtypes, and would then throw an exception again when the server was shutdown because the hooks weren't there to be removed.

Changed to explicitly cast NPCs to what they should be, and verified it's working all the way through.  There is one issue with a stable master not taking a ticket, but I think that's a larger problem with stable masters in general.


tegstewart:
I'm going to take a closer look at a bunch of quests that don't appear to be working properly, and bundle them all together.

PR #79, opened 2017-11-08
Fixed allow_cross_realm_item

tegstewart:
CheckAbilityToUseItem() was using allow_cross_realm_item to bypass item.realm = player.realm checks, but wasn't working for weapons and armour because of the way it determined which ability to check for.

Now, when non-helm armour or a weapon are checked, if allow_cross_realm_item is true, the ability checked is selected based on the player's realm rather than the item's so weapons and armour are actually cross realm now.


Graveen:
Indeed :)

PR #80, opened 2017-11-09
Fix mob stats

tegstewart:
AutoSetStats() was only being called if all a mobs stats in npctemplate or mob were 0, which led to funkiness like mobs with 0 con and 1 hp that died if sneezed at.

AutoSetStat() is now always called, and check each stat individually to see if it should be set to default, and checks against the original template when doing so in case Level changes.  An offset was added to the stat per level calculations so that a level 1 mob starts with a default physical stats at 30 rather than 31.

Also noticed that GameNPC constructors are rarely called, so I put a comment in about that to warn future devs, and set the two functions that load the NPC to call AutoSetStat(), otherwise it never gets executed.

### DIFF
--- a/GameServer/gameobjects/GameNPC.cs
+++ b/GameServer/gameobjects/GameNPC.cs
@@ -199,10 +199,7 @@ namespace DOL.GS
 			{
 				base.Level = value;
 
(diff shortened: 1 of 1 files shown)


Graveen:
Excellent, thank you

PR #81, opened 2017-11-11
Add pet spell scaling

tegstewart:
Added another optional feature to scale pet spells linearly according to their level.  Spell value and damage are scaled to resemble similar player spells.

This will make pets less powerful early on; the enchanter level 7 pet will DD for 19 damage instead of 100.  However, their spell will increase in power up to 118 damage.

The feature requires SCALE_PET_SPELLS to be set to true to work, and uses SCALE_PET_SPELLS_PERCENTAGE to allow users to select how powerful they want pet spells to be in relation to player spells.


OntheDOL:
do you have any data of pet hitting for 100 dmg? I was curious, so had a look and my level 7 pet hits for 43 dmg against a lvl 1 grey, and 33 against blue con. 
Anyhow , could this logic not go somewhere else like in gamepet itself

tegstewart:
Some of the spells seem to scale already, but some don't.  Things like procs, and dd spells seems to scale, but healing, damage adds, and damage shields definitely don't.  A high level pet's damage shield being at 2 DPS is just sad.

I could modify GameLiving.CastSpell() to take a spell argument, which would allow me to pass a modified spell to it from within GamePet.  I could also move the scaling code itself into a function like GameLiving.ScaleSpell() which doesn't do anything but could then be overloaded for GamePet, GameNPC, etc.

I'm pondering removing the SCALE_PET_SPELLS_PERCENTAGE property and replacing it with a pet effectiveness one.  This would allow both pet melee and spells to be adjusted in one place, since pets are hitting very hard right now.

tegstewart:
SpellHandler.CapPetDamage() limits damage DD spells do, but only gets called as part of CalculateDamageBase().  It doesn't seem to be called for procs, damage shields, or damage adds, and it has no impact on things like healing spells, so the underhill ally heals for around 70 all the way from the level you get them to level 50.  So your level 7 companion won't do 100 damage, but once it hits level 27 the base damage of the spell is lower than the damage cap, and the pet's DDs stop getting better.

With SCALE_PET_SPELLS set to true and the percentage at 75, pet spells are very close to player spells at level 50 in terms of both damage and healing spells.

OntheDOL:
kinda feels like a bandaid fix to me. The companions spell not scaling past level 27 is because of its value in the DB. The problem with the heal spell is valid. im not sure why you couldnt override CastSpell from inside Gamepet class. Then perhaps clone the spell in there so you can modify the spell.value based on level. Or even modifying it here in gamenpc `if (line.KeyName == GlobalSpellsLines.Mob_Spells)
			{
				// NPC spells will get the level equal to their caster
				spellToCast = (Spell)spell.Clone();
				spellToCast.Level = Level;
			}`

tegstewart:
I can't just overload the method without also changing GameLiving, but I'll find a way to do it.  I think I want to drop the scale percentage bit in favour of one impacting pet effectiveness since that allows both melee and spell scaling.

I could change it to scale assuming the spell damage/value in the DB is the desired value at level 50, but most DB releases already have damage/value lower than they should be.  It would make turning the feature on require a fair bit of DB work as well, and I'd rather this be as simple as possible for people who want to use it.

tegstewart:
I went a little down the rabbit hole with this change.  During testing, I found that effectiveness wasn't impacting melee damage at all.  GameLiving was using a hardcoded effectiveness of 1.0, so I changed it to instead use the owner's effectiveness instead.

OntheDOL:
i like that the logic was moved to gamepet, but what i dont like is you are hardcoding damage values for spells. Yes, if i choose to enable PET_SCALE_SPELLS, spell gets scaled, but its being scaled against a hardcoded value and not what i choose from my DB

tegstewart:
I don't see a good solution to that.  With a player spell, I could track it back to linexspell and get the level so I can get the damage or value to level ratio, but there isn't a way to do that for pets insofar as I can see.

I could add another server property such as pet_scale_spells_effectiveness that adjusts the spell damage and values as a percentage of player spells.  It might be useful having it as a separate setting than pet effectiveness overall.  Alternatively, I could replace pet effectiveness with pet_scale_spells, and let people adjust pet melee damage using the pet_autoset server settings.

tegstewart:
I had a chat with ontheDOL about this, and there is some scaling that is done on live that is not done in DOL.  Before dabbling with customized stuff, I'll fix that.

I think code also has to be moved into spellhandlers, as things like proc effects can't be modified when adjusting the proc itself.

Closing this PR while I redo things.

PR #82, opened 2017-11-11
Changed duraction<1 exception to log entry

tegstewart:
This exception causes weird server behaviour like mobs vanishing and such, so this changes it to log a warning instead.


Graveen:
Can't harm :D

PR #83, opened 2017-11-12
Dreaded seals

tegstewart:
Added a loot generator and NPC object to drop and collect dreaded seals.  Added four new server properties to allow drop rates to be adjusted.


Graveen:
Seems fine :)

PR #86, opened 2017-11-13
Removed hardcoded efficiency values

tegstewart:
Removed hardcoded efficiency values in functions that should have been using the Efficiency accessor instead.  Modified multiplier to effectiveness to not assume base effectiveness is 1.

This will cause those two lines to properly use overloaded Efficiency accessors from inheriting classes, and causes melee damage of pets to be properly effected by res sickness.

Also removed special characters that always pop up as a change so that doesn't happen in future.

### DIFF
--- a/GameServer/gameobjects/GameLiving.cs
+++ b/GameServer/gameobjects/GameLiving.cs
@@ -863,7 +863,8 @@ namespace DOL.GS
 		/// <returns></returns>
 		public virtual double AttackDamage(InventoryItem weapon)
 		{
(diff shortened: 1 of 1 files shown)


Graveen:
do you own a server @tegstewart  ? We could push it to supported ofc.

Graveen:
Yay ! :)

tegstewart:
Nope, no server for me.  I just play solo on my own server.

I figure anything I come up with that makes the game better for me could also make it better for anyone else, so sharing it just makes sense.

PR #87, opened 2017-11-15
Scale pet spells

tegstewart:
On live, pets spells change in quality as the pet changes levels.  As an example, an Underhill Ally will heal for more at level 50 than at level 25.

DOL doesn't currently do this.  Damage and damage shields specifically are capped by level, but heals and other effects like haste buffs aren't.  So an underhill ally heals for the same amount when you get them at level 12 as they do when you're level 50.

This remedies that, but will break older DB builds which give pets weaker spells.  PET_SCALE_SPELL_MAX_LEVEL is 0 by default to disable this feature, and DBs that are designed to use it can then assign it a value appropriate to the spells they give pets.

To mirror live, PET_SCALE_SPELL_MAX_LEVEL would be set to 44, and pets would be given spells as if they were level 44.  A level 44 underhill ally with a 200 hp heal spell would heal for 200 hp at level 44, but only 100 hp at level 22 and 45 hp when you first get them at level 12.

Another benefit to this is that pets under the effect of juggernaut have their spells improve accordingly.  An Underhill Ally at level 70 because of juggernaut will heal for 159% the amount they heal for normally.

A few other changes:  I had to give write access to spell damage, value, and duration, and I changed the heal spell handler functions to calculate everything as doubles and then round to int at the end to make calculations more accurate.


tegstewart:
I neglected to mention that this scales melee style effects as well.  A pet can be given a 5 second stun style, and with max level set to 44 it will only stun for 2.5s at level 22.

Taunt spells are scaled, but taunt/detaunt styles aren't as they already scale with damage.

PR #89, opened 2017-11-19
WorldMgr.Getobjects to Generic

LRFalk01:
Convert WorldMgr.Getobjects methods to generics. This removes casting currently being done, and it also removes reflection method building for OfType and ToArray. Should be slightly more efficient and is less code.


LRFalk01:
Closing PR as too deviant

PR #90, opened 2017-11-19
Mob archery speed fix

tegstewart:
Changed AttackSpeed() to use the ArcherySpeed rather than CastingSpeed property to reduce ranged attack speed.


tegstewart:
Removing special characters in comments is Travis breaking.  Interesting.

Graveen:
Yeah strange those chars issues. Fixed at hand by keeping master texts.

PR #91, opened 2017-11-19
Gameserver world clean

LRFalk01:


LRFalk01:
crap. didnt mean to do this ><

PR #92, opened 2017-11-20
Pet cancel buffs

tegstewart:
This adds a new method, StripOwnerBuffs(), to GamePet, which is called when the pet dies and when a player releases it.

It will stop players from summoning one pet to get buffs, and then releasing it to summon something else while still keeping the buff.


Graveen:
Brilliant !

PR #93, opened 2017-11-21
Speed up server start time

LRFalk01:
Load keeps faster by pooling their load.
Quit doing a db call per mob loading to get ambient texts.


tegstewart:
Nice!

OntheDOL:
i love me some optimization

Graveen:
Thank you ! :)

PR #94, opened 2017-11-21
Changed Respawn Time and RP/BP rewards for PvE Servers

tegstewart:
Since keeps/towers in PvE servers are farmed for seals, having the lords and mission masters respawn immediately is a problem.  This sets them to respawn according to LORD_RP_WORTH_SECONDS and sets their BPs and RPs to 0.

Ideally, PvE databases should have mob entries in keeps which have their own varied difficulties, but this is a good stopgap measure: UPDATE keep SET Realm=0; and you're good to go.

### DIFF
--- a/GameServer/gameutils/LootGeneratorDreadedSeals.cs
+++ b/GameServer/gameutils/LootGeneratorDreadedSeals.cs
@@ -1,0 +1,107 @@
+﻿/*
+ * DAWN OF LIGHT - The first free open source DAoC server emulator
+ * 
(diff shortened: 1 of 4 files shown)


Graveen:
Can you look at the conflict we have there ? TYVM

tegstewart:
Pretty sure it's because I was using an old fork that didn't have LootGeneratorDreadedSeals in it.

I'll close this and redo it from an up to date fork.

PR #97, opened 2017-11-23
Stop pets from getting aggro out of LOS

tegstewart:
I reported this a while back, but it doesn't look like it was remedied: http://www.dolserver.net/viewtopic.php?f=5&t=23087&p=158981&hilit=pet+aggro+through+wall

This is a quick fix rather than resolving the underlying problem.  There may be other issues stemming from the underlying race condition Leo mentions, but fixing that is beyond my capabilities.

### DIFF
--- a/GameServer/ai/brain/StandardMobBrain.cs
+++ b/GameServer/ai/brain/StandardMobBrain.cs
@@ -248,7 +248,10 @@ namespace DOL.AI.Brain
 
 				if (CalculateAggroLevelToTarget(npc) > 0)
 				{
-					AddToAggroList(npc, (npc.Level + 1) << 1);
+					if (npc is GamePet) // Required to stop pets from being aggro'd without LOS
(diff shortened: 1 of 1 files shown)


Graveen:
Can't agree more !

OntheDOL:
Where does this leave charmed pets? Im not sure they fall under GamePet

tegstewart:
They probably don't.  Might have to check for ControlledNPC as well.

Graveen:
I also don't think they. But such a fix is simple and should prevent a lot of bad aggro.

OntheDOL:
They both use icontrolledbrain I believe so one check for that should cover both

tegstewart:
Did a little testing, and my mentalist can't charm humanoids or undead, so I can't test this.

Methinks this just became one of  my famous "I just wanted to fix this one thing" PRs.  :-p

PR #99, opened 2017-11-25
Auto set mob & pet stats

tegstewart:
This changes several things.

First, it sets SpellHandler to use pet INT to scale spell damage the same way it does with mobs.
Second, it adds pet and mob autoset properties for INT.  It didn't make sense to use INT for mobs but not autoset it, while giving the option to pets allows the DB to control pet spell damage.
Third, I changed pet stat code to use the NPCTemplate stat values as a percentage to scale individual pet stats.  This gives the DB significant control over specific pets ie. By setting STR to 50 in NPCTemplate, you can cut a non-melee pet's STR in half, while a harder hitting pet could have STR at 125 to increase its STR by 25%.
Lastly, I tweaked autoset values for pets, as pets were getting far too much STR and as a result doing a lot more damage than they should.

The majority of DB releases have pet NPCTemplate stats set to 0, which is ignored, so it shouldn't cause problems for most DBs.


Graveen:
I like this change :D

tegstewart:
To give credit where credit is due, the pet stat scaling thing wasn't my idea.  Tolakram implemented it for Con, but only for values between 1 and 29.  I just stripped away the original functionality and replaced it with an expanded version of that.

PR #101, opened 2017-11-25
Fix 2h damage to match live

tegstewart:
Only Albion 2H and Polearm weapons should get the Spec*0.5% damage bonus as they have to dual spec.  Midgard and Hibernia get a flat 10% bonus to 2H weapon damage.

GameLiving.MakeAttack() does properly check for 2H and polearm specs, and bases their damage variance off the secondary spec, but that only impacts variance.  Essentially, Albion 2H and polearms are paying the penalty for dual specialization, but the benefit is being given to Mid and Hib 2H users who don't have to dual specialize.

This remedies that by only applying the Spec*0.5% bonus to Albion 2H and polearm weapons.

There has been some argument about that being how it works on live, so here are some sources:
http://camelotherald.wikia.com/wiki/Large_Weapons
http://camelotherald.wikia.com/wiki/Polearm

From the polearm article: "Bonus damage in compare to one hand weapons: 10% + 0.5% * [[Polearm]"  The article for Albion 2H lists a similar formula, but it doesn't appear for any of the other realms weapon skills.

It's surprisingly hard to get old grab bags in a searchable format.  This seems pertinent, but isn't 100% exact either: http://web.archive.org/web/20021020050909/http://www.camelotherald.com:80/more/474.php
"Q: When you spec in polearms, do you have to spec in a specific weapon type? 
A: It�s a good idea just to max out your average damage � speccing in polearm raises your maximum damage, and the specific weapon spec raises the bar on the minimum level of damage you will do. "

There was also a discussion of this on our own forums a little while back, but I'm not sure what was done as a result: http://www.dolserver.net/viewtopic.php?f=47&t=13343&p=96921&hilit=dual+spec+polearm#p96315


OntheDOL:
sorry, this is wrong. Alb dual spec for variance has been known since day one. Both of the credible sources for melee damage formulas state it the same way DOL has it implemented. 
You're basing your argument on "Why does Alb have to dual spec for something both hib and mid get for free?" and not basing it on research done by alot of people in the past.

tegstewart:
I'm basing it on what Mythic said shortly after release regarding how specs worked.  They've consistently said that Albion 2H and polearm spec increases max damage while the secondary damage type reduces min damage variance.

OntheDOL:
so show some evidence how HIb two hand works and mid two hand works

tegstewart:
I already did that.

We're not going to resolve this between ourselves.  That's the reason I put the PR up; we need Graveen and some of the others to look at it.

tegstewart:
Part of the challenge is that the sources that talked about this are either no longer available, or are archived in the Wayback Machine so they're very hard to search through.

Here's another one: http://web.archive.org/web/20071025125011/http://www.classesofcamelot.com/misc/skills.asp?type=Weapons
"The ability to use two-handed swords of all kinds. Training in the Two Handed Skill results in an increased overall damage and gives you access to the combat styles."
"The ability to use polearm weapons of all kinds. Training in the Polearm Skill results in an increased overall damage and gives you access to the combat styles."
"The ability to use Celtic spears of all kinds. Celtic spears are two-handed weapons. Training in the Celtic Spear Skill results in a decreased damage variance and gives you access to the combat styles."
"The ability to use two-handed weapons of all kinds. Training in the Large Weapons Skill results in a decreased damage variance and gives you access to the combat styles."

Note the difference in phrasing.  Two-handed and Polearms "increased overall damage" while Celtic Spear and Large Weapons "decreased damage variance."

OntheDOL:
how about you go on live and do some proper research and logs if you want to make a big change based on "wording" . There has been some people who have done this live research, and thats what formulas are based on. To dispute that, you should have some logs of your own dont you think?

tegstewart:
Why don't we wait for a third pair of eyes?  We're both stuck in on this issue, so we need third parties to look at it from the outside.

I've spent several hours digging into this, and I haven't come across anybody backing up either conclusion with logs.  If you've got them, feel free to post them here so everybody can see them.

OntheDOL:
you've spent several hours searching for "wording" on something that needs "testing"

tegstewart:
Here's another one: http://web.archive.org/web/20071031001213/http://www.classesofcamelot.com/faq.asp?mode=view&cat=9

"What are advanced weapons?
     Advanced weapons are Polearms and Two-Handed weapons from Albion that require a base proficiency. Because Polearms and Two-Handed weapons encompass all three damage types - piercing, slashing, and crushing, you need to be proficient in one of the base weapon types to become better at an Advanced Weapon.

Umm, I don't get it.
     O.K. Here's how they work: First you need to choose a base proficiency - piercing, slashing, or crushing. Second, you need to commit yourself to being almost completely offensive in combat. 

Remember how I said training in Weapon Skills decreased your damage variance? If not, go back and that page of the FAQ. That's not true for Polearms and Two Handed weapons. When you train in these Advanced Weapons, your variance is determined by the highest base proficiency you have - be it piercing, slashing or crushing. If you only train in an advanced weapon, you will get access to all of the combat styles for the advanced weaponry, but your damage variance will never decrease. Also, unlike any other weapon skill, training in advanced weapons increases your maximum damage! 

So, pick a base proficiency, train in it to decrease your variance, and then train in the advanced weapon to open up your combat styles and increase your maximum damage.

Seems to me like that requires a LOT of skill points?
     You're right, it does. But you'll out damage all other weapon types (at least, that's what's supposed to happen.) I've heard stories of melee characters that use advanced weapons doing up to 50% more damage in combat than those who don't use advanced weapons. This is why I say you must commit yourself to being completely offensive - you will not have enough skill points left over after training in a base and an advanced weapon to try for shields or something else of that nature, although I would highly recommend you do not slack off too bad in your parry ability.

Can you re-cap the pros and cons for me?
     Sure thing. The pros of advanced weaponry are that you will be a beast in melee combat and do hordes of damage and have really cool advanced weapon styles. The cons are that it takes a lot of skill points, so you will be very focused, and you will need to train in a base proficiency to reduce your damage variance."

tegstewart:
ontheDOL, I'm not going to discuss this with you further until we have somebody else looking at this.  At this point, us discussing it is not productive.

OntheDOL:
It's something that  should have been 'discussed'  somewhere like the forum rather PR based on nothing but indirect wording and assumptions, meh

tegstewart:
Pendragon tests done by Tolakram: http://www.dolserver.net/viewtopic.php?f=50&t=15709&hilit=albion

This throws a monkey wrench into the mix, as he's finding that when fully spec'd both Mid and Alb 2H users are hitting for 147% of 1H damage.  That's a lot higher than the current code provides.

tegstewart:
Closing this after a discussion with Leo.  Tolakram's testing reveals that, whatever Mythic said about how it worked, Alb don't get any kind of bonus damage.

dol-leodagan:
For record here is my input : 

@tegstewart I made some research, and I agree with @ontheDOL, Albion doesn't have any kind of special bonus for Two-handed dual-spec'ed melee damage.
Tolakram tested it in 2010 : http://www.dolserver.net/viewtopic.php?f=47&t=13343&start=60

I think the purpose of dual-spec for Two-Handed in Albion is  to nerf the ability to use any type of damage from their 2H specs.

There are similar mechanics for other realm, Mid have free 2H ability but lacks a thrust damage line, Hib have Large Weapon for Crush/Slash, and Celtic Spears for Thrust damage and they don't receive any 1H ability for spec'ing in 2H weapon...

Alb is just being hindered from having such great 2H specs, and dual-spec'ing is not in pure loss, you also receive 1H skills and damage improvement from the mandatory points in there, leaving room for some 2H/Shield switch in Combat...

From my research Alb could also be balanced by having superior styles in their 2H lines, I read some player reports stating that growth rate could be higher compared to Mid 1H/2H lines, Endurance consumption could be lower, etc etc...

Anyway absolutely no evidence that there is a bonus in the base damage calc method !

For reference here is the most documented damage calculation formula actually implemented (maybe with errors) in DOL : http://camelotherald.wikia.com/wiki/Melee_Damage

PR #103, opened 2017-11-28
Fix npc left handed swing

tegstewart:
GamePets were hard coded to never left hand swing, and Necromancer pets were hard coded for some pets to always swing with both hands.  I removed both, so the Underhill friend can LHS as on live now.

There was a bunch of checks before getting calling CalculateLeftHandSwingCount(), which were repeated further down.  However, the second time they didn't check for GameNPCs, leading to them not being able to LHS.  I removed the duplicate checks, and verified that GamePets with LHS chance attack with their off hand.

I didn't change anything in GamePlayer, which overrides all these anyway, but I tested a Mercenary just to make sure dual wield is still working properly.


Graveen:
Excellent code maintenance. Thank you !

PR #104, opened 2017-11-28
Fixed hangabout theurgist pets

tegstewart:
On the next Think() after a theurgist pet's target dies, it dies as well rather than standing around until its duration runs out.


tegstewart:
I thought that was how live behaved.

I'll change it back right away.

PR #106, opened 2017-11-30
Heal threshold server props

tegstewart:
Removes hard coded percentages at which NPCs heal.  Separated into one server property for mobs and pets and one for keeps as they were using different hardcoded values.


Graveen:
Kudos !!!

PR #107, opened 2017-12-02
Redid LootGeneratorDragonscales

tegstewart:
LootGeneratorDragonScales looks like it was originally copied from the Aurulite or Atlantean Glass generators, and was dropping huge numbers of dragon scales inappropriately.

This redoes the logic to drop fewer scales while taking the target and target's level into account.  Because it checks for named dragons, it can be used instead of adding scales to their loot template in the DB.


Graveen:
Pesky tabs ;)

tegstewart:
Pesky really is the best adjective for them.  :-p

PR #108, opened 2017-12-05
Server prop for Alb weapon dual spec

tegstewart:
This is an old and controversial issue.

At release, Mythic said that Alb weapons did more damage than other realms, at the cost of having to spend more spec points to get there, but tests on live disproves that.  I remember Hib tanks being pissed because once they maxed out parry and spear, they had wasted spec points, while Alb tanks still had useful things to spend points on.  There's still people trying to get Broadsword to change it on live.

Since live still dual specs, the responsible option is a server prop set to dual spec by default.  Let people decide for themselves if they want dual speccing or not.

### DIFF
--- a/GameServer/gameobjects/GameLiving.cs
+++ b/GameServer/gameobjects/GameLiving.cs
@@ -1745,24 +1745,23 @@ namespace DOL.GS
 					weaponTypeToUse.Object_Type = weapon.Object_Type;
 					weaponTypeToUse.SlotPosition = weapon.SlotPosition;
 
-					if ((this is GamePlayer) && Realm == eRealm.Albion && ServerProperties.Properties.ENABLE_ALB_WEAPON_DUAL_SPEC)
+					if ((this is GamePlayer) 
+						&& (GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.TwoHandedWeapon) 
+						|| GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.PolearmWeapon))
+						&& ServerProperties.Properties.ENABLE_ALB_WEAPON_DUAL_SPEC)
 					{
 						// Albion dual spec penalty, which sets minimum damage to the base damage spec
-
-						if (GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.TwoHandedWeapon) || GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.PolearmWeapon))
+						if (weapon.Type_Damage == (int)eDamageType.Crush)
 						{
-							if (weapon.Type_Damage == (int)eDamageType.Crush)
-							{
-								weaponTypeToUse.Object_Type = (int)eObjectType.CrushingWeapon;
-							}
-							else if (weapon.Type_Damage == (int)eDamageType.Slash)
-							{
-								weaponTypeToUse.Object_Type = (int)eObjectType.SlashingWeapon;
-							}
-							else
-							{
-								weaponTypeToUse.Object_Type = (int)eObjectType.ThrustWeapon;
-							}
+							weaponTypeToUse.Object_Type = (int)eObjectType.CrushingWeapon;
+						}
+						else if (weapon.Type_Damage == (int)eDamageType.Slash)
+						{
+							weaponTypeToUse.Object_Type = (int)eObjectType.SlashingWeapon;
+						}
+						else
+						{
+							weaponTypeToUse.Object_Type = (int)eObjectType.ThrustWeapon;
 						}
 					}
 				}
--- a/GameServer/gameobjects/GameLiving.cs
+++ b/GameServer/gameobjects/GameLiving.cs
@@ -1748,7 +1748,7 @@ namespace DOL.GS
 					if ((this is GamePlayer) && Realm == eRealm.Albion
 						&& (GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.TwoHandedWeapon) 
 						|| GameServer.ServerRules.IsObjectTypesEqual((eObjectType)weapon.Object_Type, eObjectType.PolearmWeapon))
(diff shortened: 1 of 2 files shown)


OntheDOL:
I dont really see the point to having a 17 year game mechanic as an 'option'.
There really is no end in sight when any constant / game mechanic can potentially become a server property. 
Just my 2c, but i don't like the direction DOL would be heading with this

dol-leodagan:
I don't like it either, but Albion's two-handed weapons mechanism is already an exception to the "normal" rules, the patch is written as to disable this exception when the property is configured by user, looks clean and is not completely custom behavior... (it only skips some part of the damage computation)

I don't like the name of the property, "dual spec" can be mistaken with dual wield and similar dual weapon specializations, "alb" and "spec" should be written in full length "albion" and "specialization" for clarity, you should maybe refer to "Albion Heavy/Advanced Weapon Specialization", and the description could be more redacted. (use verbs and literal description of what the variable does)

You shouldn't have removed the Realm comparison, in Player Versus Environment server (PvE, Coop Rules) other realms members could equip Polearms or Albion's Two Handed Swords / Hammers (probably Hero using Spears or Large Weapon Specialization, maybe Midgard fighters with Hammer or Sword ?), so without filtering this rule to Albion's Classes only other realms will always end up having lowest damage for these weapons (except maybe if the proficiency with other realm's weapons is mapped elsewhere in PvE Ruleset, but it would greatly hinder Hibernia's Classes who haven't specialized in base weapon)

To sum up, I think this is a good addition to DOL, the implementation logic is simple enough and does not clutter the code, but this rule should still apply only to Albion's members, and naming/phrasing need some more work 👍

tegstewart:
The realm comparison was redundant as the two weapons specs which are then checked for are only available to Albion classes.  It's akin to testing to see if the player is an Alb, and then checking to see if they're an Armsman or Paladin.

In my mind, DOL is an opportunity to fix the mistakes that Mythic made and Broadsword hasn't remedied.  We can make DAOC better than livelike, and there's already precedents for that.  The quickcast icon appearing in buffs and the 25% run speed boost out of combat are both DOL specific features that are better than livelike.

Leo, it's probably better if you update the server property yourself rather than me changing it to fit your description.

benfalk:
> At release, Mythic said that Alb weapons did more damage than other realms, at the cost of having to spend more spec points to get there

Where is this claim at?

OntheDOL:
the 25% run speed bonus out of combat is actually taken from live ;)

tegstewart:
@benfalk I listed a bunch of different sources for things here: https://github.com/Dawn-of-Light/DOLSharp/pull/101

tegstewart:
@OntheDOL The description when it was added made it sound like it was custom, but I'm entirely prepared to be wrong on that.

benfalk:
@tegstewart I took a gander and didn't find any statement officially from Mythic; perhaps I missed it, but I only found 3rd party sites detailing it.  Was it in a grab bag for instance that I missed?  I'm just genuinely curious if this is on official statement Mythic ever put out and did nothing about or conjecture from the community.

tegstewart:
@benfalk Part of the problem is that Mythic didn't archive their old notes.  I spent a lot of time digging through the internet archive trying to find it, but that removes search functionality so you have to do it one article at a time.  I remember the controversy on release, but of course that's entirely anecdotal, and not really pertinent anyway since live doesn't appear to have ever behaved that way.

tegstewart:
To be clear, this is a bee in my bonnet, but that doesn't mean that it has to go into DOL.  I changed it for my own edification, and am a big believer in making things I change available to others since there's likely somebody else out there who doesn't like it either.  If the community doesn't want to add this, that's fine.  I'm not going to take it personally either way, and I can always change it in my own server.

dol-leodagan:
@tegstewart it's great that you made the change regarding my input, I'm glad you tested the PvE Rules to make sure it wasn't adding a buggy behavior into this particular setup !

For the other comments, I would just like to point this patch is not adding really more complexity to the damage rule there, it's just end up being a switch that activate or deactivate a specific Albion's rule in the code, and whatever are the players/officials reports on Albion's Polearm/2H damage this does not add an arbitrary bonus to the rule, it just make Polearm and Two-Handed behave like other Realm Specializations.

This could be a customization that improve gameplay on shards that don't want to strictly match official server rules (for example a Coop Server won't be unbalanced with this...)

If we disregard wether this is official or not, is there any drawbacks to merge this PR ?

OntheDOL:
I guess there is no real limit to what can be a server property. 
ALLOW_MALE_VALKYRIES
USE_CUSTOM_FALL_DAMAGE
ACUITY_BUFF_AFFECTS_HYBRIDS_AND_SUPPORTS
REMOVE_SPELL_FOCUS
all sarcasm aside, i guess there is no drawback

tegstewart:
@OntheDOL Funny you mention that.  I've been pondering putting in a server property to have acuity buffs work on pets, now that they actually have an acuity stat to buff.  :-p

OntheDOL:
hehe as i posted, it did cross my mind i just gave you your next 4 PR's 😉

tegstewart:
To be fair, I try to make changes on the DB side before I dive into code.

A good example of that is my Necro heal over time reminder.  I tweaked it to also cast the spell on the shade, with 0 healing value and a longer duration so it starts flashing 3s before the actual heal ends so you know when to recast it.

dol-leodagan:
@OntheDOL 
Even if some of your remarks are being sarcasm, there are update that I made toward allowing customization of classes/races combo, but it's through GameServerScript because it was obviously hard to describe this kind of constraints dictionary in a property. (for example newer revision of the game allowed new character selection options)

Not being nit picky but even the other values you mentioned could be customizable, acuity behavior is probably hardcoded in core instead of being customizable through a scripted SpellHandler, fall damage is probably a "const" somewhere in an obscure source file, and for the last one about Focus I hope this can be handled through a database edit of the spell table...

But there is tons of other things we try to make customizable when enhancing the project (custom classes, custom skills, custom specialization tree etc...) and for values that are buried in the Core logic that can't be easily subclasses we have to rely on some global variable.

tegstewart:
@dol-leodagan I wonder if a lot of that could be moved back out and into the DB or into scripts in a more elegant fashion than having it all hardcoded in core.

Graveen:
Ok, it does not harm, so let's go :)

PR #113, opened 2017-12-10
Fixed faulty "fully healed" messages

tegstewart:
Because amount is a double which is being rounded to an int, sometimes the amount actually healed rounds down, and you end up with the amount to heal being more than the amount actually healed, which triggers the "fully healed" message when it isn't appropriate.


Graveen:
Brilliant !

PR #114, opened 2017-12-19
Added Parameterised Database.SelectObject() methods

tegstewart:
The current SelectObject() methods are flagged as obsolete, but no parameterized versions exist for them.  This uses the existing parameterized SelectObjects() methods to get lists, and then return the first object in them.


tegstewart:
Closing this as I rolled it into PR #116

PR #116, opened 2017-12-19
Scale bd minion effects by player spec

tegstewart:
This change makes low spec BD pets useful without making them overpowered for the amount the player has invested in them.  It also provides more reason to branch out BD specs and to gear for +skill bonuses.

Currently, pet spell scaling is based entirely on pet level.  If you summon a bone guardian, it will cast the same quality of buffs if you only have 20 spec in suppression as if you spec'd all the way to 50.  That's just not right, and leads to dipping into a secondary spec being more powerful than it should be.  A good example would be a BA BD speccing to 15 in suppression to get the same full damage shield and attack speed buffs they'd get if they were specced to 50 in suppression.  There should be a consequence to that which also provides incentive to spec further into suppression and to gear +suppression skill items.

By capping the pet caster level when scaling to the player's modified spec, low level minions are still useful but not disproportionately so.  It also provides incentive to gear for secondary spec bonuses.  With full gear and RR bonuses at 20 spec, it almost reaches parity: 20 + 11 for gear + 10 for RR = level 41 when 42 is the max level for minions on live.

Even the melee pets are impacted by this as their style effects are scaled, so the bleed and melee speed reduction from pillager/plunderer will vary according to player spec.  The only subpet it won't impact is archers, and if archers are changed to use archery spells down the road they will be impacted as well.

### DIFF
--- a/DOLDatabase/IObjectDatabase.cs
+++ b/DOLDatabase/IObjectDatabase.cs
@@ -93,6 +93,36 @@ namespace DOL.Database
 		
 		#region Select Where Clause With Parameters
 		/// <summary>
+ 		/// Retrieve a Single DataObject from the database based on the Where Expression and Parameters Collection
+ 		/// </summary>
+ 		/// <typeparam name="TObject"></typeparam>
+ 		/// <param name="whereExpression"></param>
+ 		/// <param name="parameters"></param>
+ 		/// <returns></returns>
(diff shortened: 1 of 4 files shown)


tegstewart:
Oh, right.  My PR for adding GetSingleObject() methods needs to go through before this can.

tegstewart:
Rolled the parameterised SelectObject() methods in rather than having them as two separate PRs

PR #117, opened 2017-12-19
Fixed issue with archer mobs at max range

tegstewart:
Sometimes, mobs will try to fire arrows but due to some gimmick in the math somewhere be just a hair out of range.  This leads to them starting to fire over and over again without actually firing.  It appears to add multiple ranged fire timers, so the moment the target is in range, all the timers go off at once and do a ridiculous amount of burst damage.

By forcing mobs to move just a tiny bit closer before firing, this issue appears to be resolved.


tegstewart:
Of course, after thoroughly testing this, I had it happen again just now.  So I've got a little more work to do on this.

PR #119, opened 2017-12-20
autoselect caster for realm spells

tegstewart:
This adds caster autoselection for beneficial spells.  Let's say your valkyrie is soloing a mob, and you need to cast an insta-heal.  In live implementation, you have to target yourself, cast the heal, then target the mob.  With this change, the heal spell will automatically target the caster when the spell isn't appropriate for the current target, so you can just heal yourself without having to change targets.

While this isn't a live feature, it feels like a no brainer to me.  Other MMOs from the same generation included this as a feature, and it's ridiculous that live doesn't.  That said, it's really easy to add a server property to enable this if the community wishes.

### DIFF
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -346,6 +346,19 @@ namespace DOL.GS.Spells
 		#endregion
 
 		/// <summary>
+		/// Sets the target of the spell to the caster for beneficial effects when not selecting a valid target
+		///		ie. You're in the middle of a fight with a mob and want to heal yourself.  Rather than having to switch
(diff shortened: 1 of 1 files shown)


Graveen:
I agree with you but i think this must be ServerProperty'sed.
Mainly because tricky actions can be part of the gameplay and therefore participate in class balancing.
In example, what is the need for timed selfs buffs like champion or CL ones that last 20 minutes is questionnable, or why some spells are movecast and some aren't, etc... I think this can be done "on purpose".

tegstewart:
Can do.  Gimme five minutes.  :-)

tegstewart:
Hmm.  May have to close this and redo it as a branch.  Hold on.

PR #120, opened 2017-12-21
Lord calls for help

tegstewart:
This causes keep lords to call for help when attacked, pulling every surviving guard in the keep to come help them without aggroing a specific target so they attack the first hostile they see.

Once we have pathing working, this will allow things like bottlenecking at stairwells, but for now guards just charge in from all directions so killing guards off before tagging the lord is a really good idea.

I know this is how Gaheris behaves, but I'm not sure about normal live.  Do keep lords exhibit the same behaviour on normal live servers?


tegstewart:
This is getting a little messy, and I think I have a better way to do it anyway, so I'm going to close this and redo it.

PR #124, opened 2017-12-24
Report keepitem duplicates & Optimization

tegstewart:
There is a lot of pattern matching in here, so AppVeyor won't like it.

This started out with me wanting to put in logging to see if I could track down why dolpub is putting multiple lords and other characters in keeps.  I thought the code for GameKeepComponent.FillPositions() was difficult to red and maintain, so I changed it to make it more legible.

While I was there, I figured I may as well optimize code a bit by doing things like using .Add(), Contains(), and Remove() rather than [] on Hashtables, and from there looked for other things I could speed up.  Ultimately, I ended up changing Hashtables to Dictionaries to force proper typing to catch coding errors, and speed things up by removing the boxing and unboxing necessary with Hashtables.

I also found a bunch of small things like key mismatches in which an item was added using one thing as a key but another when removing it, and in some cases a string/int key being used to add but another type used when trying to remove.  So I fixed those.

This produces a LOT of warning log entries.  I'm not sure if that's due to mistakes in dolpub that weren't being reported before and thus weren't ever addressed, or if I'm reporting errors when it's not appropriate.  I'll be working on cleaning up the DB next, and if I find some of those log entries are inappropriate I'll come back and remove them from code.

### DIFF
--- a/GameServer/commands/gmcommands/keep.cs
+++ b/GameServer/commands/gmcommands/keep.cs
@@ -2166,7 +2166,7 @@ namespace DOL.GS.Commands
 
 							d.Component = new GameKeepComponent();
 							d.Component.AbstractKeep = k;
-							d.Component.AbstractKeep.Doors[d.DoorID] = this;
+							d.Component.AbstractKeep.Doors.Add(d.DoorID.ToString(), d);
 
 							d.Health = d.MaxHealth;
 							d.StartHealthRegeneration();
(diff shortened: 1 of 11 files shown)


Graveen:
AppVeyor dislike 🗡

LRFalk01:
He used pattern matching which is a c#7 feature.

tegstewart:
It's a C# 7 thing.  Rather than doing this:
GameLiving living = obj as GameLiving;
if (living != null)
{...}

I'm doing this:
if (obj is GameLiving living)
{...}

Given the number of case statements which all start by checking object types, it reduced line count significantly while making code easier to read.

PR #125, opened 2017-12-24
Server props for keep guard stats

tegstewart:
Stats for keep guards and lord was hardcoded in TemplateManager.  This replaces that with server properties for all guard and lord stats.

The default values are based on the old hard coded values so that guard/lord stats won't change on existing servers.


Graveen:
Thank you :)

PR #126, opened 2017-12-25
Give DB more control over pets

tegstewart:
These changes allow the brain of pets to be set by the ClassType in their NPCTemplate, and allows the number of subpets BD commanders can summon to be set by the TetherRange of their NPCTemplate.

Not only does this allow databases to add pets like the Dread Lord on live, but it also allows custom servers to create completely new pet behaviour without recompiling by changing which brain pets use or even creating and assigning a completely new scripted brain.

Amusingly enough, formations for a BD commander with 5 subpets are exactly like live; lines get longer and the last two pets stand over the commander for other formations.

Old hardcoded values are still present, so existing servers should not be impacted.


Graveen:
Excellent !

PR #127, opened 2017-12-28
Fix Housing Teleporters

LRFalk01:
Currently you can buy a teleporter to any realm. This is not a good thing. Fixed this unless server type is PvP or PvE.


Graveen:
Thank you :)

PR #128, opened 2017-12-28
Speed up DB init

LRFalk01:
Parallel registering of db objects on server start.


Graveen:
You have an estimation of the speedup ? :)

LRFalk01:
15-30 seconds

PR #133, opened 2018-02-04
Set default value in db for number types

LRFalk01:
Currently there are a bunch of fields in the DAL defined as being nullable, but the datatype in C# forbids this.  This changeset corrects this for number types and for MySQL sets the default value to 0 correcting any existing db data. I did notice that there are still DateTime fields marked as nullable.


Graveen:
Thank you :)

PR #136, opened 2018-05-12
Resurrection illness and Realm Speed restored

OntheDOL:
Saves the effects of resurrection illness and Speed of the Realm on player exit and restores them when player logs back in.

Had to move these effects to a different 'globalspellline' as reserved spells are not saved on exit.


OntheDOL:
Just a mental note here for me, if this goes through I need to add a spellline to the public DB

Graveen:
ty mate

PR #140, opened 2018-06-28
Add Consignment Merchant

OntheDOL:
allows player to purchase a consignment merchant from the housing lot marker after the lot is purchased (along with house type and porch etc) Mainly for convenience.


Graveen:
:)

PR #142, opened 2018-06-28
Make Lord heal self-only

tegstewart:
Currently, Lords have a heal spell which has realm as the target, so they can heal other mobs.  Since there's no LOS check on that, it leads to lords spamming heal over and over again to heal guards on the outside of the keep.

This changes the three lord heal spells to be self targeted instead.


Graveen:
Thank you !

PR #143, opened 2018-06-28
Sort npc spells

tegstewart:
Sorts spells into categorised lists, which will make AI modifications a lot easier and more efficient.

I cleaned up and added null checks to the style list code as well.


Graveen:
@tegstewart thank you. Is the IsHealing stuff breaking the actual database compatibility with various existing DBs ?

tegstewart:
It doesn't change the DB at all, so it shouldn't.

PR #145, opened 2018-07-02
Added checks for spell effects already on the target

tegstewart:
Brains weren't checking for spells like heals over time at all, and weren't checking if damage over time spells were already on the target, causing it to spam damage over times spells over and over again.

When casting a spell with a duration, it will now check to make sure the target doesn't have already have it.


Graveen:
Thx !

PR #148, opened 2018-07-07
Add dreaded seal collector.cs

tegstewart:
DreadedSealCollector.cs was written and present, but not added to the solution.


Graveen:
Not correct mate ?

PR #151, opened 2018-07-08
Fix Stacking Buffs, added more buff types

tegstewart:
Added a few more buff types that mobs and pets weren't checking for.

LivingHasEffect() was checking for both matching spell effect group and spell type, which is incorrect.  Offensive procs should stack, as should pulsed damage adds and damage buffs.  Spell effect group exists specifically to allow those combinations, so it's odd to not use it exclusively.


Graveen:
let's go ! :)

PR #152, opened 2018-07-08
 Make PvE Lord heals interruptible

tegstewart:
Removes the uninterruptible tag from PvE lords' self heal.  Lords should be soloable in PvE, and an uninterruptible self heal makes that impossible.


Graveen:
Or having an SP with heal between time relative to players in the area ? It would scale smoothly.

PR #153, opened 2018-07-18
Fix archer bug

tegstewart:
Causes archers to drop aggro on targets that move outside their aggro range.  This prevents them from continuing to draw and fire arrows each Think(), leading to all the arrows striking the target at once when they're back in range.


Graveen:
Good move !

PR #155, opened 2018-07-27
Characterization tests

NetDwarf:
This PR is meant to be a start at making the code base less legacy. Modus operandi: create characterization tests meaning I just check the results of a certain setup. I of course do a bit of white box testing to avoid unnecessary tests. Characterization tests are a kind of unit test in this context.

Unit tests have to be quick (usually millisecond rather than seconds), isolated and may at no time make calls to outside dependencies like database or file system. Therefore the current UnitTests are mostly integration tests. They are important as well, but they are not a replacement for unit tests, but instead complement each other. 

Unit tests should be the first thing to run before anything is changed and they should be all green. Of course with legacy code you need to check what they mean to you and the naming I explain below should aid in this question.

Why all that? I think everyone can see that the code base in its current form is not ideal. Global state, temporal coupling, classes with dozens of responsibilities reaching thousands of lines, null checks everywhere as well as type checks and castings. It is obvious how this could happen. You make a change, but you don't want to change the structure because it might break everything or something you didnt thought of. The more you add the less confident you can be about what might break and the less you change the worse it gets. To make necessary refactorings more viable unit tests are usually the weapon of choice.

Naming:
- namespace is DOL.UnitTests.*
- classes and file names begin with UT_ (so you can see at all time which is the UT and which is the actual class in your IDE)
- tests are named like this: UnitOfWork_Scenario_ExpectedOutcome where UnitOfWork is usually the tested method
- names may be long (try to be concise anyway) because unit tests are never called and the names are meant as a documentation of the observed behavior (!)
- tests are structured with Arrange Act Assert in mind

Possible caveats with this PR (I am aware of): 
- NSubstitute as new dependency
- DBSpell constructor might actually make calls to the DB(?)
- Travis/Appveyor might not work anymore, because of the change in the first commit (please test!)


dol-leodagan:
There are some functional changes in your commits, you should move them to another PR (GameNPC, GamePet, GamePlayer changes)

And ideally Continuous Integration Tests should succeed unless there are some dependencies that can only be resolved after merge ?

NetDwarf:
Thanks for the answer :)

Changes in these classes are as following:
GamePlayer: - changed CharacterClass Accessor to virtual
GameNPC: - parameterize constructor (safe refactoring)
GamePet: - added the parameterized constructor from GameNPC (super class)

There is no functional difference and that was kind of the point. These changes are only implemented to make the code testable.

However the reason why the CI tests fail is that I changed the project structure of the tests. The former unit tests were renamed to IntegrationTests and DOLDatabaseTests were merged into IntegrationTests. Maybe it would have been better to make a separate PR, but it would not make too much sense without actual unit tests. Also the actual unit tests are running isolated so it does not make a difference for the structural change to the integration tests. The unit tests run regardless of your system setup (the project should compile of course).

PR #156, opened 2018-07-30
Characterization tests

NetDwarf:
This PR is meant to be a start at making the code base less legacy. Modus operandi: create characterization tests meaning I just check the results of a certain setup. I of course do a bit of white box testing to avoid unnecessary tests. Characterization tests are a kind of unit test in this context.

Unit tests have to be quick (usually millisecond rather than seconds), isolated and may at no time make calls to outside dependencies like database or file system. Therefore the current UnitTests are mostly integration tests. They are important as well, but they are not a replacement for unit tests, but instead complement each other.

Unit tests should be the first thing to run before anything is changed and they should be all green. Of course with legacy code you need to check what they mean to you and the naming I explain below should aid in this question.

Why all that? I think everyone can see that the code base in its current form is not ideal. Global state, temporal coupling, classes with dozens of responsibilities reaching thousands of lines, null checks everywhere as well as type checks and castings. It is obvious how this could happen. You make a change, but you don't want to change the structure because it might break everything or something you didnt thought of. The more you add the less confident you can be about what might break and the less you change the worse it gets. To make necessary refactorings more viable unit tests are usually the weapon of choice.

Naming:
- namespace is DOL.UnitTests.*
- classes and file names are named  "UT_<class>" (so you can see at all time which is the UT and which is the actual class in your IDE)
- tests are named like this: UnitOfWork_Scenario_ExpectedOutcome where UnitOfWork is usually the tested method
- names may be long (try to be concise anyway) because unit tests are never called and the names are meant as a documentation of the observed behavior (!)
- tests are structured with Arrange Act Assert in mind

Possible caveats with this PR (I am aware of):
- NSubstitute as new dependency
- DBSpell constructor possibly makes calls to DB


dol-leodagan:
You're adding 3 new nuget dependencies :

- Castle.Core
- System.Threading.Tasks.Extensions
- NSubstitute

Could you give a short summary of what they are for ?

I don't know if DOL should stay with as low dependency as it is now, but I think it doesn't matter much if these are only used for tests. (to make sure the build/run is always as easy as possible for our contributors)

NetDwarf:
Castle.Core and System.Threading.Tasks.Extensions are dependencies of NSubstitute. NSubstitute is a Mocking-Framework. It is the one with the best readibility in my opinion. You just create stubs with `var fakedClass = Substitute.For<Interface/Abstract>();` and you can manipulate that instantiated stub the way you need it and it acts like the real thing. You can also use it for actual classes however properties/methods need to be virtual.
For stub behavior: `fakedClass.Method().Returns(value);`
For mock behavior: `fakedClass.Received().Method(arguments);`

The data collection in the Microsoft license with System.Threading.Tasts.Extensions is the biggest problem I have with it. 

The next best imo would be FakeItEasy with anonymous lambda functions. Not as readable and a bit slower, but it does the job. It has no further dependencies beyond itself.

dol-leodagan:
Thanks for the explanations.

I didn't understand the issue with System.Threading.Tasts.Extensions, but FakeItEasy sounds like it could be a good replacement in case of trouble.

It's a bit annoying that NSubstitute forces class to have virtual methods or properties, does FakeItEasy have the same requirement ?

Anyway this look good enough that it could be merged as is, let me know if you have anymore update !

NetDwarf:
To answer the question directly: FakeItEasy has the same requirement.

I think only JustMock and TypeMock can fake behavior on non-virtual methods. It is a smell tho. Like you shouldn't test private methods/properties/members which is possible with some hackery. JustMock/TypeMock can do that too iirc. The most obvious disadvantage of both frameworks is, that they cost money, other than that I don't know much about them beside that :p

Usually the only concrete class you need in a test is the class which is under test in TDD. Which is of course not the case with legacy code, because necessary interfaces are missing and would be awfully long, because classes have often dozens if not hundreds of responsibilties :D

NetDwarf:
I will add a few tests I made when I was refactoring Stats in GamePlayer. The refactoring result was not to my liking, but the tests are still valuable.

NetDwarf:
Forgot to say, that my PR is done with the last commits I made.

dol-leodagan:
I think this will need more fine tuning, but for now nobody contribute to Tests so it's a start !

I don't like seeing test methods in GamePlayer (CreateTestableGameplayer static methods...)

PR #158, opened 2018-08-08
Bard heal song fix

OntheDOL:
Fix for bards healing song (regrowth spec) being removed from 'song' list when songs from 'Nurture spec)' are specced into.

Moved the index variable outside of the foreach loop. Problem is that the bard has two spec lines with songs (unlike the other song classes) . This leads to the index being reset to -1 twice instead of only once the first loop ( seemingly how the logic was intended) 
Renamed 'index' to 'songIndex' just for a little clarity on this specific case

Tested all the song classes and songs still being added/ replaced fine and bard heal song now fixed

PS. messed up original commit as i forgot to put it in a branch so there is some janky commit history to be able to recommit the changes


dol-leodagan:
Ok I'll merge this !

dol-leodagan:
@Graveen Well I thought he already had them...

I invited him to DOL organization but I think I forgot to put him in a team ;)

PR #159, opened 2018-09-21
Start Level >1 Health/Mana/Endurance fix

Loki1976:
Ensure player starts with full Health/Mana/Endurance when level >1 in server properties using played time <1.


Graveen:
Yeah for sure Thx mate :)

PR #160, opened 2019-02-15
new emotes

OntheDOL:
some emotes that got added a while back but missing from DOL. '/Cower' exists but i couldn't find the ID for now.


Graveen:
Thank you VM :)

PR #162, opened 2019-05-30
Fix debuff in SpellRangePercentCalculator

NetDwarf:
Change (int) rounding for debuff calculation to be applied after multiplication


NetDwarf:
Short explanation: NearsightReduction is supposed to be reducing the effectiveness of Nearsight. 10(%) NearsightReduction and a 50(%) SpellRange debuff (Nearsight?) would effectively be a 45(%) SpellRangeDebuff

Before: Any value of NearsightReduction > 0 would yield a 0% effectiveness of debuffs because of the (int) cast.

PS.: Found this while refactoring the whole bonus system. The tests for this are added in the working branch and will be PRed once the bonus system is in good shape :)

Graveen:
Seems totally legit. Thank you !

PR #164, opened 2019-06-15
Added DreadedSealCollector.cs and LootGeneratorDreadedSeals.cs

tegstewart:
The files were already there, they just needed to be added to the project.


Graveen:
Thank you :)

PR #165, opened 2019-06-16
Fix exploitable side effect of stat (de)buffs

NetDwarf:
Remove GamePet conditional in PropertyChangingSpell.OnEffectStart because it added a second bonus but never removed it


NetDwarf:
Stat buffs (including AF buffs) on GamePets added a second bonus where the Value was determined by the Spell.Value (the base) and the Spell.Level. For example Constitution added ArmorAbsorption bonus and all magic resists additionally. 

The biggest problem however is, that these additional bonuses are not removed when the buff is removed, so you can for example add/remove constitution a few times and suddenly have magic resist cap on your GamePet without any buffs.

I removed it without replacement because I think that most of these second bonuses are not livelike (and MeleeSpeed didn't even work). For the resist portion I am sure. The melee absorption (aka ArmorAbsorption) is as far as I know only determined by constitution and maybe dexterity, but not by AF. If it is dependend on AF this can be added in the ArmorAbsorptionCalculator instead of PropertyChangingSpell

tegstewart:
My understanding was that the secondary bonuses were supposed to apply to pets, but they weren't supposed to get the primary bonuses.  This seems to be backed up by this grab bag: https://darkageofcamelot.com/article/its-friday-grab-bag-waitwhat-day-it

The problem appears to be that pets are getting the buffs applied in OnEffectStart but they aren't being removed in OnEffectExpires.

NetDwarf:
That is generally true, but Strength does not increase the MeleeDamage bonus, but the Strength bonus. Strength in turn increases effective melee damage (in GameLiving.MakeAttack, ouch). The same applies for ArmorAbsorption and constitution and probably also dexterity. Here the ArmorAbsorption is calculated in the corresponding Calculator. On live no Stat buff increases magic resists and pets are very susceptible to magic damage from players.

There is a point to be made, when it comes to the factors that translates constitution and dexterity to ArmorAbsorption. This is skewed now ofc and I could adjust it, so that it caps the same as before.

I am currently refactoring the Bonus and Property system and I found this bug while routinely sanity checking that I haven't broken something or missed a unit test case.

tegstewart:
There still needs to be some NPC vs player specific code in the mix, if only because stats scale differently.  Off the top of my head, the server properties for NPC strength values add strength at 10x the rate of other stats.

I'm really curious about your bonus and property system overhaul.  Will that be resolving this issue inherently anyway?

NetDwarf:
Strength was never really that effective on pets and I would guess that it is the same calculation as for players. Con and probably dex however determine melee absorb instead of hit points. This is implemented in ArmorAbsorptionCalculator. I could change the factor for determining this, as the absorption calculation felt pretty livelike. 

Here a big problem is obvious: there is no differentiation between bonuses and properties. All are calculated in the calculators and some calculators are more of a bonus calculation and some are more of a properties calculation and some are in the middle. 

**Regarding the "overhaul":**

I don't want to change behavior at all. I created new classes for bonuses (and properties is the next) and delegate old methods/classes to use these instead and deprecated some of them. You can see the progress [here](https://github.com/NetDwarf/DOLSharp/tree/BonusRefactoring).

[This](https://github.com/NetDwarf/DOLSharp/blob/BonusRefactoring/GameServer/propertycalc/PropertyIndexer.cs) is one the key abstractions from the old data type (deprecated PropertyIndexer clients). It's also way more memory efficient. Optimization is not a focus at this point however. I'm still working to make it simpler and more readable.

Graveen:
So the code you removed where applyed twice in fact ? 
I mean, buffs/debuffs of pets are still working ? 
Thank you !

NetDwarf:
@Graveen  a second (and different) bonus was added in addition to the actual bonus and this bonus was never removed so it was stacked ad nauseam if you applied->removed buffs over and over. This was probably to make the buffs behave more livelike, but it would be advised to change the factor in the Calculator (where for example Constitution and Dexterity on NPCs contribute to ArmorAbsorption). 

Tested this, because I couldn't believe it, and it really works exactly like that.

If Endless Conquest is available I will make a comprehensive test on live.

Graveen:
Thank you !

tegstewart:
I confirmed it as well by chain casting pet class buffs and watching the benefits stack via /mob info.

PR #166, opened 2019-06-18
Now drops variable seals based on keep level

tegstewart:
Number of seals dropped is now based on keep level
BGs only drop 1 seal
Keep lords and Agramon now properly drop sanguine rather than glowing seals
Removed code for dragons since they aren't in regions that will ever use this loot generator anyway
Mucked about with whitespace, with mixed results


Graveen:
Seems fine ! TY

PR #167, opened 2019-06-19
Added con loss, xp loss, ress sickness on death server properties

tegstewart:
Added server properties to replace hard coded values determining at which level characters start to lose xp on death, lose con on death, and suffer ress sickness on death.

I did it to set those values at level 51 and not have them happen at all, but figured some folks might want to still have such things at high levels.


Graveen:
Excellent, TY

PR #169, opened 2019-06-20
Replaced hard coded keep guard respawn time with server properties

tegstewart:
Keep guards were hard coded to respawn between 5 and 25 minutes.  This replaces that with two server properties: GUARD_RESPAWN and GUARD_RESPAWN_VARIANCE.

This is something I wanted for PvE servers, since keep lords pull the entire keep when aggro'd, so having guards respawn before you get to the lord makes keeps nearly impossible.  I figured it was better to expose this for people to be able to tweak it rather than hard code a different value for PvE/PvP servers.


Graveen:
TY, good move

PR #171, opened 2019-06-27
Add artifact quest placeholders

tegstewart:
There were a bunch of placeholder artifact encounters without matching artifact quests.  I added placeholder quests to make those artifacts available.

They were also using two hard coded strings for ArtifactID and Init().  I changed it to use a common property, and moved hard coded properties to the beginning of the file to make them more visible.


tegstewart:
Odd.  I'll close and retry.

PR #172, opened 2019-06-27
Add place holder artifact quests

tegstewart:
There were a bunch of artifact encounters which lacked associated quests, making those artifacts unavailable.  I added placeholders to make those artifacts available.

I noticed in existing artifact quests that ArtifactID and Init() were using identical hard coded strings, so I replaced that with a new m_artifactID property and moved hard coded properties and accessors to the beginning of the files to make them more visible.


Graveen:
TY !

PR #174, opened 2019-06-30
Added FREE_RESPEC server property

tegstewart:
Adds a server property making respecs free.

I added an option to respec champion abilities at trainer when the property is set.  I could make it general case if people would like.


Graveen:
Thank you !

PR #175, opened 2020-03-10
Hostile Effects trigger GameNPC.OnAttackedByEnemy()

tegstewart:
Only damaging spells and spells that were resisted were calling OnAttackedByEnemy().  As a result other spell types could be used to pull and bypass bring a friend checks.

### DIFF
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -2725,6 +2725,10 @@ namespace DOL.GS.Spells
 				ad.IsSpellResisted = false;
 
 				m_lastAttackData = ad;
+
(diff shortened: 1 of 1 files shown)


Graveen:
Thank you. Note, on live, insta debuffs are not triggering aggro.

tegstewart:
The new BAF code might make this change superfluous.

My concern was that OnAttackedByEnemy() only does anything if a mob isn't in attack state.  That led to situations in which it wasn't calling BAF code as the mob went into attack state before OnAttackedByEnemy() was called.

OnAttackedByEnemy() also adds the attacker to the aggro list and starts the attack, so I thought it was better to fix other possible issues rather than just fixing BAFing.

However, if things like debuffs aren't supposed to trigger an immediate response, this probably isn't desirable.

PR #176, opened 2020-03-10
New baf

tegstewart:
The old bring a friend code used two separate mechanisms, one of which was only used in dungeons.  The new code is used everywhere, is more consistent with live, and adds two server properties to allow the odds of mobs bringing friends to be adjusted.

Mobs don't always trigger a AttackedByEnemy event before going hostile, which was leading to BAF code not always being called.  I've moved the BringFriends() call to AddToAggroList() instead to ensure it is always called.

The new implementation fixes the following situations in which mobs wouldn't always BAF properly:
Mob attacked on its own
Pulled by pet
Pulled by non damaging spell such as debuff
Mob BAFing multiple times in a fight
Mobs brought by a friend also trying to BAF

### DIFF
--- a/GameServer/ai/brain/ControlledNpcBrain.cs
+++ b/GameServer/ai/brain/ControlledNpcBrain.cs
@@ -1029,7 +1029,7 @@ namespace DOL.AI.Brain
 			AttackMostWanted();
 		}
 
-		protected override void BringFriends(AttackData ad)
+		protected override void BringFriends(GameLiving trigger)
 		{
 			// don't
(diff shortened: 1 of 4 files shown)


Graveen:
TY, seems good. 

On live, insta debuffs are not triggering aggro (ie a champion can debuffs before attacking).
I think this is the same for casted debuffs (although not sure - perhaps it splits between resist debuffs and stats debuffs, idk).

But your changes are welcomes so it is merged :)

PR #177, opened 2020-03-10
Update OnlineStatusUpdate.cs

Loki1976:
Use Server Username and Password from DB for updating server on portal.


Graveen:
This is an excellent change, as i think IF there were people still using XML the server is not connected to the portal.
Anyway, perhaps it is throwing a NRE with XML ?

PR #178, opened 2020-03-12
Added allow_train_anywhere functionality

tegstewart:
I really like Phoenix letting people use the /train command to train anywhere in the world rather than having to run back to their trainer, so I added that functionality.  Since it's not livelike, it's disabled by default and has to be enabled by the allow_train_anywhere server property.


tegstewart:
I just logged into live, and found that /train brings up a trainer window on live.

Should I leave this as disabled by default, or remove the server property and have /train bring up a trainer window all the time?

Xanxicar:
Just set the default value to enabled imo.
It's nice to have on/off switches for patch level dependent features.

tegstewart:
Done.

PR #179, opened 2020-03-13
Necro pet spellcasting can be interrupted

tegstewart:
Necro pets weren't being interrupted when spellcasting.  I added code to ensure the pet can actually cast spells before adding them to the queue, and used OnAttackedByEnemy() to check for interrupts and cancel spellcasting/clear the spell queue.

That meant that instant spells could be queued behind ones with cast times and be lost if the former spell was interrupted and the queue cleared, so I had instant spells bypass the queue and cast immediately.


tegstewart:
I've been doing a lot more reading up on necros, and my implementation might be off.  Necro pets should be able to start casting during combat, but can be interrupted once they start casting and should also be interrupted when they swing.

I'm going to have to put some more work into this, so I'm closing it for now.

PR #181, opened 2020-03-14
Remove double pet heal scaling

tegstewart:
Pet heals are already scaled in SpellHandler, so pet heals were being scaled twice.


tegstewart:
This appears to be a bigger problem.  Some spell handlers don't call base methods and assume they will only ever be cast by players but are cast by pets on live, leading to possible null exceptions.

I'll close this, and PR a solution which fixes it across the board.

PR #182, opened 2020-03-14
Added BAF_MOBS_ATTACK_RANDOM_PLAYERS server property, expanded to check battlegroups

tegstewart:
Live almost always sends BAF mobs after the player who pulled, so I added a server property to choose whether mobs go after random players or after the player who pulled them.

I also expanded the code to look for battlegroups so that it counts and sends mobs after all the players in a raid rather than just the players in the pulling group.  Since BGs can be up to 64 players by default, I tried to make the code as efficient as possible.

During testing, I found that BattleGroup.GetPlayersInTheBattleGroup() always returned an empty array, so I fixed it.

### DIFF
--- a/GameServer/ai/brain/StandardMobBrain.cs
+++ b/GameServer/ai/brain/StandardMobBrain.cs
@@ -920,12 +920,12 @@ namespace DOL.AI.Brain
 		/// <param name="attacker">Whoever triggered the BAF</param>
 		protected virtual void BringFriends(GameLiving attacker)
 		{
-			if (!m_canBAF)
+			if (!canBAF)
 				return;
 
 			GamePlayer trigger;	// player that triggered the BAF
 
-			// Only add on players and pets of players
+			// Only BAF on players and pets of players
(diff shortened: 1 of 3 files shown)


tegstewart:
There is some debate about how BAF code works on live, so we should hold off on committing this until live functionality is confirmed.

Graveen:
Interesting, TY

Graveen:
too bad i just went to commit. Anyway i think this is way better than it was so i suggest to not revert.

tegstewart:
Nalfgar was telling me that his testing found mobs attack seemingly random group members but don't attack BG members.

Which kinda makes sense.  If some bonehead in a BG pulls a boss, it's better if the boss kills their group rather than wiping the entire BG.

We could still use nearby BG members to determine numbers of adds, but send them after the pulling group.

That's really the challenge.  How do we want this to work?

Graveen:
I think BG encounters are very specifics, and need their own brains / logics / workflows. So i would let classic mobs in BAF code focus on group only.

PR #184, opened 2020-03-15
Improved pet spell scaling

tegstewart:
Pet spell scaling code is now significantly better.  Rather than scaling spells as pet cast them, spells are scaled as they're added to the pet.  Subspells and style effects are scaled on the fly by SpellHandler and StyleProcessor as they aren't stored on the pet.

There is no longer any need for spell scaling in specific spell handlers.

A fringe benefit is that spells that necromancer pets cast automatically and their style effects will be scaled, but spells cast by the necromancer won't.

### DIFF
--- a/GameServer/gameobjects/GameNPC.cs
+++ b/GameServer/gameobjects/GameNPC.cs
@@ -3406,7 +3406,7 @@ namespace DOL.GS
 		/// <summary>
 		/// Sorts styles by type for more efficient style selection later
 		/// </summary>
-		protected void SortStyles()
+		public void SortStyles()
 		{
 			if (m_stylesAnyPos != null)
 				m_stylesAnyPos.Clear();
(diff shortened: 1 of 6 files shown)


Graveen:
This is a big change, i let it open for comments

tegstewart:
It's disabled by a server property by default, so it shouldn't impact most servers.  That said, I should talk to LRFalk about changing Eve to use it.

Right now the level 24 BD heal minion's heal spell has a value of 70, and the level 48 one's value is 150.  Meanwhile, the enchanter heal pet's heal has value 70 all the way from when you get them at level 12 until you're level 50.  Doesn't seem right.

tegstewart:
I should clarify: This is a better implementation of something I did back in 2017: https://github.com/Dawn-of-Light/DOLSharp/pull/87

Graveen:
ok ;)

PR #187, opened 2020-03-16
Exposed necro pet stats via server properties

tegstewart:
This is the same thing I did with mob, pet, and keep guard stats.  Greater necroservants have different stat calculations, so I had to use a second set of server properties for them.


Graveen:
Thank you !

PR #188, opened 2020-03-16
Update OnlineStatusUpdate.cs

Loki1976:
Use Server Username and Password from DB for updating server on portal


Loki1976:
wrong button

PR #191, opened 2020-03-18
Updated pet buff stripping code

tegstewart:
StripOwnerBuffs() had a potential exploit; a player could join a group, be buffed by a groupmate's pet, leave the group, and not have buffs cast by the pet stripped.

Pets now track which targets they've buffed via AddBuffedTarget(GameLiving).  StripOwnerBuffs(owner) has been changed to StripBuffs(), since the owner is no longer needed.

I added a few other things along the way.  Spells now have IsHelpful and IsBuff properties, IsHealing and IsHarmful are now more efficient, and SpellHandler.HasPositiveEffect now calls Spell.IsHelpful rather than duplicating code.


Graveen:
Very interesting, TY

PR #192, opened 2020-03-19
Added submenus and many more options

tegstewart:
I significantly expanded the PvE/PvP teleporter, using minimalist formatting and submenus to give access to a whole bunch of new places.  I added New User Journey towns, dragon zones, classic and SI dungeons, Darkness Falls, and Agramon.

Since this is the PvE/PvP teleporter, there shouldn't be any balance issues stemming from reducing travel time.

Most of the new teleport destinations weren't in Eve, so I put in a PR adding them.


tegstewart:
Redoing this.

PR #196, opened 2020-03-22
Fixed BringFriends() bugs/exploit, added more server properties

tegstewart:
Added several new server properties to expose BAF functionality for BGs.  Changed constants to virtual accessors so they can be properly overridden, and added an accessor to specify range for the player IsWithinRadius() check.

By default, BGs are ignored, number of mobs added is based on number of nearby group members, and adds attack random nearby members of the puller's group.

I tested this exhaustively and found a few bugs and possible exploits.  Code now checks for duplicates, players who are in both the group and the BG so they aren't counted twice, but only when absolutely necessary and as efficiently as possible.


Graveen:
TY !

PR #200, opened 2020-03-26
Give crafting skill points as GM.

Loki1976:
Wanted to test some crafting features so created this. 

/player craftskillpoints <Skill> <Amount>


Loki1976:
Seems I've re-invented the wheel with this as we have /crafting add which I didn't even notice :/

Closing

PR #201, opened 2020-03-27
Server reboot command for GM's.

Loki1976:
Ability to reboot server for GM's that may not be in front of machine.


Graveen:
Thx !

PR #202, opened 2020-04-07
Improved pet spellcasting AI

tegstewart:
This is a significant overhaul of pet AI, which took me much further down the rabbit hole than I expected.  Since pet AI is far simpler and easier to test than mobs, I've tried to change as little of StandardMobBrain as possible.  I plan to work on that next, but it will be a separate PR.

Changes:
Pets now use sorted spell lists rather than stepping through all of their spells each tick.
Pets will now stop melee attacking to cast offensive and defensive spells so long as they aren't being interrupted.
Pets now use instant spells when in melee and in the middle of casting other spells.
Pets that can heal now check for in two passes to prioritize healing badly injured targets, and now heal their owner before themselves.
Added more robust spell checking to catch invalid spell casts earlier to prevent things like attempts to heal targets which are out of range or use HoTs that are already on the target.
NPCs now load abilities, styles, spells, block/parry/evade chance, and left hand swing chance from npcTemplate even if the template is set to not replace mob values, as those aren't in the mob table so mobs weren't getting them otherwise.  This will impact 1100+ npcTemplates in Eve, so a lot of mobs are suddenly going to have spells/styles/abilities/etc they were supposed to have before.
GamePet.CanCastInCombat() no longer always returns true, stopping pets from casting offensive spells while simultaneously attacking in melee.

The biggest change was making GameLiving.CastSpell() and it's overrides return a bool to indicate whether the spell started to cast successfully or not.  Previously, brains had to assume spells were going to cast, leading to odd bugs in which the spell didn't cast, but the mob would stop attacking as if it had.  The most complicated change is GamePlayer.CastSpell(), where it likely won't ever be used, but c'est la vie.

### DIFF
--- a/GameServer/ai/brain/ControlledNpcBrain.cs
+++ b/GameServer/ai/brain/ControlledNpcBrain.cs
@@ -384,7 +384,7 @@ namespace DOL.AI.Brain
 		public override void Think()
 		{
 			GamePlayer playerowner = GetPlayerOwner();
-			
+
 			long lastUpdate;
 			if (!playerowner.Client.GameObjectUpdateArray.TryGetValue(new Tuple<ushort, ushort>(Body.CurrentRegionID, (ushort)Body.ObjectID), out lastUpdate))
 				lastUpdate = 0;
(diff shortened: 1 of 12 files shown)


Graveen:
Let's try this... Thank you for the hard work.

PR #204, opened 2020-04-09
Added /player levelup GM command

tegstewart:
Something useful I've been using to test pets leveling up with their owners.


Graveen:
Thank you !

PR #206, opened 2020-04-13
Optional: Pets level with owner

tegstewart:
This is another relatively simple sounding thing that led to a lot of changes.

Pets now level up with their owners if the PET_LEVELS_WITH_OWNER server property is true, default is false.  Pet spell scaling was revisited to have spells re-scaled when the pets level changes.

BD subpet spell scaling being capped by spec is now enabled by the PET_CAP_BD_MINION_SPELL_SCALING_BY_SPEC server property, disabled by default.  BDSubPet overrides pet scaling and sorting functions to move BDSubpet specific code out of GamePet.

Code to calculate pet level was moved from SummonSpellHandler to GamePet.  GamePet now stores the damage and value from the spell that summons them to allow their level to be changed on the fly.

Fixed a bug with Necro pet spellcasting interruptions only happening when a second spell was queued.

Mob DB entries, npcTemplates, and server properties for stats are now shorts to reduce the number of casts from int required.

AutosetStats() was being called multiple times unnecessarily.  There should now be a lot fewer calls, and they should be more efficient.

ControlledNpcBrain's GetLivingOwner(), GetNPCOwner(), and GetPlayerOwner() used two different implementations which were run back to back when GetLivingOwner() was called.  I've changed it to use a single simpler implementation which only runs once.

BDSubpets and Necro pets had special code increasing their levels based on patch 1.87.  Those changes should be done in the DB rather than in code; I'll be submitting those changes to Eve shortly.  Necro pets now scale stats by npcTemplate like other pets, and log a warning if the npcTemplates have stats set to 30 as in older DBs.

I changed a bunch of log objects from protected to private, as inherited classes were creating log entries that misrepresent the class that entered them.

### DIFF
--- a/DOLDatabase/Tables/Mob.cs
+++ b/DOLDatabase/Tables/Mob.cs
@@ -54,14 +54,14 @@ namespace DOL.Database
 		private int m_meleeDamageType;
 		private int m_respawnInterval;
 		private int m_faction;
-		private int m_Constitution;
-		private int m_Dexterity;
-		private int m_Strength;
-		private int m_Quickness;
-		private int m_Intelligence;
-		private int m_Piety;
-		private int m_Empathy;
-		private int m_Charisma;
+		private short m_Strength;
+		private short m_Constitution;
+		private short m_Dexterity;
+		private short m_Quickness;
+		private short m_Intelligence;
+		private short m_Piety;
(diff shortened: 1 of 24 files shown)


Graveen:
Ok let's continue the pet revamp, thank you.

PR #207, opened 2020-04-13
Added GetTeleportLocation() to create dynamic teleport locations

tegstewart:
Copy/pasted GetTeleportLocation() from GameTeleporter and modified it to work outside of a realm specific teleporter.

Also changed DestinationRealm parameter to destRealm to avoid confusion with the GameTeleporter DestinationRealm accessor.


Graveen:
Thank you !

PR #209, opened 2020-04-20
Finished BD commander implementation

tegstewart:
Finished implementations for some BD commander options.  I also added support for the Dread Lord, and tried to make the code work as well as possible for future pets the code doesn't recognize.

Switching weapons is a lot more robust.  Added a static cache of BD weapon templates that commanders and minions draw weapons from, retrieved when the first commander is summoned.  If the keys it looks for aren't in the DB, it creates temporary default templates instead and logs a warning.  Dread Lords don't use the cache, and unrecognized commanders will be able to switch weapons based on which weapon slots are populated by their templates; this means the weapons the template gives them will be clobbered by the ones in the cache, but they can use commands to switch weapons like most other commanders can.

Taunt is a lot closer to livelike, with melee attacks generating extra aggro based on a percentage set by a server property.  I reached the default value through testing; it's enough to consistently hold aggro against a spec BA DoT, but not enough to hold aggro against both spec and baseline DoTs.

Turning on and off whether or not minions assist commanders is implemented.  It's especially useful if your commander has a ranged attack but one of your minions doesn't; you can tell the minions to stay put and have them assist once the target gets closer rather than having them pull a Leroy Jenkins.  It's especially useful given that some ranged commanders (looking at you, archer) won't ever pull aggro from the taunting subpets at bonus taunt aggro only applies to melee weapons.

Dread guardian/lich spell selection is implemented, but only for spells which the npcTemplate gives them.  Selections which don't have an appropriate spell give an error message when selected.  They'll use DoTs, dd+debuff, and regular DD spells in that order if you give those to them.

The language file for commander pets needs to be there for commanders to work, so I copied the English version into languages that didn't have it.  It'll be the wrong language, but at least it'll work.  I tried to keep the German version as German as possible, but somebody will have to go through it and translate a few things for me.  The 1.109 client doesn't seem to like some of the letters used in German, so I replaced them when for commands as that was the only way to get it to work.

Other changes:
Uses an enum for commander type rather than doing a whole bunch of name string comparisons.

GamePet.GetDamageType() was returning elemental damage for pets without weapons for some reason, so I gave all BD pets & subpets (excluding the healer minion) weapons to match live, and had GetDamageType() call the base method so it uses the damagetype from the npctemplate.

NPCs/pets will no longer spam stun defensive styles like brutalize, and will wait until the stun effect invulnerability wears off.

DD+Debuff spells were only being cast until the spell hit, and were then being ignored because the target had the effect.  NPCs will now keep casting debuff spells so long as they also do damage.  Note that CheckOffensiveSpells() doesn't know if the npc has a regular DD as well, so it will keep casting the dd+debuff rather than switching to a regular DD.

DD+Debuff pet spells now scale both damage and duration, as some pets on live scale their debuff amount between 10-30% depending on level.

Stopped pets with self-buffs spamming "you must have a target to use this spell" and interrupting their attacks trying to cast them on other targets.

Commanders not doing anything will now respond when a minion is attacked rather than just standing there watching.

Intercept chance now checks for both GamePets and BDSubPets, as live has had SM and minion pets that intercept attacks on their summoner at one point, and pet CheckAbilities() will now use abilities on GetPlayerOwner() so subpets use abilities on their summoners rather than on the pet that controls them.


Graveen:
Thank you ! Nice work

PR #210, opened 2020-04-20
DE CommanderPet

TheRealKataract:
Translated the CommanderPet file.
Got some german bases but they are not perfect. So feel free to correct if your german is better ;D

PS : First time with Github, i hope i did it right xD


Graveen:
Seems ok :D

PR #211, opened 2020-04-20
Realm ranks 14 added

TheRealKataract:
Realmpoints needed and titles names come from official website.

PS : Always trying to understand how Github works xD


tegstewart:
The entry for GameObjects.CommanderPet.WR.Const.Empower in CommanderPet.txt needs to have the special character removed, as old versions of the client don't recognize it and it stops that command from working.  Special character outside of the commands are fine; the client still doesn't recognize them but it doesn't impact functionality.

Thanks for the help!

Graveen:
keep up the good work ;)

PR #213, opened 2020-04-21
Added 2H bonus for guards/mobs/pets

tegstewart:
Guards/mobs/pets lose their chance to block when using a 2H weapon, but do not get any bonus damage to make up for it.  This adds the lost block chance as a percentage of damage to compensate.  Server properties allow it to be turned on or off for guards, mobs, and pets separately.

Keep lords & mission masters get a 15% block chance, fighters get 10%, and healers get 5%, so this will definitely have an impact on guards but not a game breaking one, so 2H damage is enabled by default for keep guards.

There are 12 mob npcTemplates in Eve that give a block chance and a 2H weapon, and they paradoxically only have 2H weapons so they can't ever block anyway.  Since they don't lose anything it doesn't make sense to give them bonus damage, so mob 2H damage is disabled by default.

I tested BD pets on live, and it's about a 30% difference in DPS with a 2H weapon, which matches Eve's 30% top tier BD commander block chance, so 2H bonus damage is enabled by default for pets.

### DIFF
--- a/GameServer/gameobjects/GameNPC.cs
+++ b/GameServer/gameobjects/GameNPC.cs
@@ -3965,6 +3965,35 @@ namespace DOL.GS
 		}
 
 		/// <summary>
+		/// Returns the Damage this NPC does on an attack, adding 2H damage bonus if appropriate
+		/// </summary>
+		/// <param name="weapon">the weapon used for attack</param>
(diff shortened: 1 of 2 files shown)


Graveen:
Thank you for recent commits

PR #215, opened 2020-04-28
Added livelike NPC buff/debuff effects

tegstewart:
Added livelike NPC buff/debuff effects based on testing on live.  Most stats replace their usual effects, but dex is still applied as it affects casting speed and ranged weapon skill.  This also makes debuffs more effective; dex debuffs in particularly were useless prior.

Debuff effects on live are incredibly powerful, and scaling buff effects to the same degree would make pets ridiculously overpowered, so buff effects are cut in half.

Server properties control effect buff/debuff effectiveness on keep guards, mobs, and pet independently, and allow them to be scrapped altogether.  The default is based on testing debuffs on live.

Other fixes:
The magic absorption property wasn't being checked by resist calculators, so it wasn't actually doing anything.

Added melee bonus damage to /mob info and /mob stats.

Fixed debuffs not being able to go below base stats.

Debuffs are applied inconsistently, with some code adding them to properties while other removed them, so a lot of calculators now apply absolute values.

The Heretic Reanimate Corpse spell was applying the magic and armor absorption bonuses to categories the calculators don't check.

Single stat debuffs were using buff effectiveness instead of debuff effectiveness.


tegstewart:
That's an interesting problem.  Hmm.

PR #218, opened 2020-04-30
Added doppelganger class

tegstewart:
My last few PRs have been rather tedious, so this is something a lot more fun.

Doppelgangers disguise themselves as invaders until injured, when they show their true form.  At half life, they summon a pet chthonian crawler to help them, and give RP and BP when killed based on server properties.

Also added an abstract class for NPCs which summon pets to assist them when injured.  I'll have to revisit it down the road, and will likely have to add a new brain and/or class for NPC pets.

Found a potential bug with pet summoning spells accidentally limiting pet level to 1 by having value = 0.


Graveen:
Thank you !
/mob create DOL.GS.Doppelganger

PR #225, opened 2020-05-24
/mob copy creates mob again

### LINKED ISSUE #224: /mob copy command doesn't work
when i attempt to copy a mob, it simply doesn't work, nothing happens, no "Unknown mob" or anything.

NetDwarf:
Moved content of else branch in MobCommandHandler.copy() one level up, so the mob is actually created if an NPCTemplate is present

Closes #224


NetDwarf:
Fixes #225

NetDwarf:
Fixes #224

Graveen:
Thank you, faster than me :)

PR #226, opened 2020-05-24
add 1.124+ basic support

JVirant:
add 1.124+ basic support (based on Los-Ojos/DOLSharp-1126 branch) + few fixes/optimizations

fix WeakMulticastDelegate.InvokeSafe
remove virtual from Packets Read/Write methods

A big change between 1.109 and 1.124+ is the position system which is from int to float for player's positions (not handled like that in this commit, I convert everything to int) and refresh rate for updates.

It's a really basic support, there are some untested packets (especially about quests).

I don't recommend to allow 1.109 and 1.124 clients on the same server (fix your server properties), with the different refresh rate, 1.109 player movements aren't smooths for 1.124 players.

### DIFF
--- a/DOLBase/Network/PacketIn.cs
+++ b/DOLBase/Network/PacketIn.cs
@@ -45,8 +45,6 @@ namespace DOL.Network
 		{
 		}
 
-		#region IPacket Members
-
 		/// <summary>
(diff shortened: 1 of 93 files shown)


Graveen:
TY Dre

PR #232, opened 2020-05-27
fix an issue with NPCCreationRequestHandler in v1.126

JVirant:
I added code to remove npc which is not in the update range when the client do something with it


Graveen:
ty

PR #233, opened 2020-06-03
Fix Heretic single damage-only focus spells

### LINKED ISSUE #222: Heretic Focus Damage spells behavior different on live
Hello, 

I am currently poking around a bit in the Heretic spell code. First of all I want to lay out what I found out on Live so it is at least clear what has to be done.

tl;dr:
1. damage growth for each successive tick is m * ( 1 + growthrate * (n-1) ) where m is base damage and n is the number of (successive) focus ticks
2. resists do not have an influence on the damage of following ticks
3. the focus ends after a certain duration
4. growth is 50% for all damage focus spells but the uninterruptable where it is 60%
5. there is a growth cap which can not be exceeded by more successive ticks

**Current/wrong state:**
Example 1:
`1. tick: 100 Damage`
`2. tick: 125 Damage`
`3. tick: 156 Damage`
`4. tick: 195 Damage`
and so on ... so the damage is roughly m*1.25^n-1 where n is the number of ticks and m is the damage on first tick before variance.

Example 2:
`1. tick: 100 Damage`
`2. tick: Resist //lowers effective damage by 25%`
`3. tick: 93 Damage`
here we see that the addition and subtraction is not symmetric. 1*(1+0.25)*(1-0.25) ≠ 1

**Same examples on Live:** (assuming 100 base damage and 25% growth)
Example 1: 
`1. tick: 100 Damage`
`2. tick: 125 Damage`
`3. tick: 150 Damage`
`4. tick: 175 Damage`
This is more of an extrapolation of the Live behavior. There is no spell with this growth and start damage, but the pattern is the same. Damage difference is always from the delve multiplied by growth rate (on Live you see 50% or 60%)

Example 2:
`1. tick: 100 Damage`
`2. tick: Resist //implicitly 25 damage added to 125`
`3. tick: 150 Damage`
Resists do not stagnate or even lower the damage, but increase it implicitly for the next tick. The 3rd tick does 150 Damage regardless of resists if it is not resisted itself.

Example 3: (growth_cap of 60%)
`1. tick: 100 Damage`
`2. tick: Resist //implicitly 25 damage added to 125`
`3. tick: 150 Damage`
`4. tick: 160 Damage // (60%+100%)*initialDamage`
`5. tick: 160 Damage` 

Another thing to mention is duration. The focus stops after a certain duration (or number of ticks if you will) and this duration is even affected by spell duration (surprised me). 

This is not exhaustive, so I will add some additional problems as well.

NetDwarf:
Add duration to focus spells
Add configurable growth rate of focus damage (LifedrainReturn in DB)
Add (configurable) growth rate cap for focus damage (AmnesiaChance in DB)

Fixes #222

### DIFF
--- a/GameServer/GlobalConstants.cs
+++ b/GameServer/GlobalConstants.cs
@@ -1065,7 +1065,7 @@ namespace DOL.GS
 		public static bool IsExpansionEnabled(int expansion)
 		{
 			bool enabled = true;
-			foreach (string ex in ServerProperties.Properties.DISABLED_EXPANSIONS.SplitCSV(true))
+			foreach (string ex in Util.SplitCSV(ServerProperties.Properties.DISABLED_EXPANSIONS, true))
 			{
 				int exNum = 0;
 				if (int.TryParse(ex, out exNum))
--- a/GameServer/commands/admincommands/BenchmarkCommand.cs
+++ b/GameServer/commands/admincommands/BenchmarkCommand.cs
@@ -46,7 +46,7 @@ namespace DOL.GS.Commands
 			
 				case "listskills":
 					start = GameTimer.GetTickCount();
-					Enumerable.Range(0, 1000).AsParallel().ForEach(i =>
+					Util.ForEach(Enumerable.Range(0, 1000).AsParallel(), i =>
 					{
 						var tmp = client.Player.GetAllUsableSkills(true);
 					}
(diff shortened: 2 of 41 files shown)


NetDwarf:
I actually intended to fix the snare spells in the same PR, since I wanted to consolidate them. Turns out snare spells are way more borked and I need to change my approach a bit, but for now at least the damage-only focus spells are working.

LanguageMgr and Util were changed to non-static classes since static classes are pretty much not usable in test. I used this approach for it https://martinfowler.com/bliki/StaticSubstitution.html

Graveen:
I loved making SplitCSV as extension method. This was my first extension method iirc ^^
Nowadays i'd use a nuget package implementing CSV splitting :D

Thank you !

PR #239, opened 2020-07-18
Fix focus spell behavior

### LINKED ISSUE #238: Focus Spell are not properly removed on move or cancel
Two distinct bugs are in focus spells, which however look fairly similar.

When character is moving while a focus is running, the focus is interrupted and properly removed from Concentration spells however the effect is still on the target for the remaining duration which should not be the case. On Live the focus spell is immediately removed.

The second one is cancelling the spell by clicking it again, which has the same remaining duration problem, but here it also doesn't remove the EventHandler and therefore gets triggered once the character moves (long after it has been cancelled). 

I think I found a possible fix for it and will post it later or tomorrow, if I don't find an obvious bug in it.

NetDwarf:
Remove check in FocusSpellAction for pulsing spells and always cancel effect
Add FocusSpellAction call in CastSpell(GameLiving) when same focus spell is executed while still running (therefore cancelled)

Fixes #238

### DIFF
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -446,7 +446,13 @@ namespace DOL.GS.Spells
 			if (Spell.Pulse != 0 && CancelPulsingSpell(Caster, Spell.SpellType))
 			{
 				if (Spell.InstrumentRequirement == 0)
+				{
+					if (Spell.IsFocus)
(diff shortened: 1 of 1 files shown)


NetDwarf:
This would introduce a subtle bug where FocusSpellAction is constantly called while moving. I close it for now.

PR #242, opened 2020-09-04
Fix 2H damage bonus

tegstewart:
2H damage was being set rather than multiplied, leading to damage being reduced to ~1.


Graveen:
thx mate !

PR #244, opened 2020-09-06
Fix Quest Mob Selection

tegstewart:
Some quest mob names are reused (Elizabeth), and others are duplicated in strange places in Eve.  This made some quests impossible.  Quests now check for name, region, and x/y coordinates to make sure the correct NPC is being selected rather than assuming the first one in the list is the one we want.

Tweaked a few things as well, making coordinates match Eve and correcting incorrect quest text.


Graveen:
Thank you !

PR #245, opened 2020-09-07
Fix most aspects of focus damage spells

### LINKED ISSUE #234: Heretic Focus Damage Snare(+AE) not livelike
- [ ] target requirement needs to be fixed (you can click anywhere on Live)
- [ ] simultaneous double tick for all snare damage focus spells need to be fixed
- [ ] growth and cap need to be applied like on normal damage focus spells
- [ ] make power usage dependent on tick and not number of targets
- [ ] remove damage fade out on AE damage (all targets get same base damage)
- [ ] if unfocused target walks out of focus radius it should no longer be damaged

NetDwarf:
Note: Uninterruptible flag really makes the spell uninterruptible which is not intended for focus damage spells. A fix involves a bigger rework on how rupt works and I am not sure that I am able to do it.

Fixes #234


Graveen:
Thx !

PR #247, opened 2020-09-10
Add /vault command

tegstewart:
Added a command to allow the player to open their vault out in the wild.  Disabled by default.


Graveen:
yes ! 👍

PR #250, opened 2020-09-17
Fix Focus spells

### LINKED ISSUE #238: Focus Spell are not properly removed on move or cancel

NetDwarf:
Removes cancel of focus spells (caster starts spell cast anew)
Focus spells should now immediately stop when concentration is broken
Fixes a nasty bug where events are subscribed to, but never removed. The events are stacked ad nauseam (each tick) and can always be resurrected by casting, moving etc. ( [This is a good description of the behavior](https://ladimolnar.com/2015/09/14/the-weak-event-pattern-is-dangerous/)

Fixes #238


Graveen:
Very interesting :) thx for the link :)

PR #254, opened 2020-09-21
Better Item Craft Button Enable Code

tegstewart:
The craft button for items was only enabled if the player's primary craft skill was spellcasting or alchemy, even if the item wasn't player crafted and thus couldn't be modified.  This meant that servers which allow all crafting skills wouldn't ever enable the button.

The item packet code now enables the craft button when the item is player crafted, the item is a type that can be imbued or tinctured, and the player has either alchemy or spellcrafting.


Graveen:
Good, thank you for the code optimization (GetCraftingSkill)

PR #256, opened 2020-09-23
More Spellcrafting Fixes

tegstewart:
Removed the redundant "You don't have enough skill to combine items" error that was popping up when an item is first opened in a self crafting window.

Fixed a null exception when working with self crafting windows, as trade.Partner is null in that case.

Kind of hilarious given that I found these while trying to replicate an issue that turned out to be Storm specific.  8-)


Graveen:
Thanks ! :)

PR #257, opened 2020-09-24
Fix Naked Script NPCs

NetDwarf:
Fixes Dawn-of-Light/Storm-issues#1


NetDwarf:
Most readable and narrow patch I could do to fix script NPC equipment loading. This code segment is so convoluted that it is pretty much edit and pray.

Graveen:
Thanks ! 👍

PR #258, opened 2020-10-05
Fix DataQuest with no Money/XP Reward

NetDwarf:


NetDwarf:
(Characterization) Tests are a lot broader than necessary for the fix. They were created during refactoring, but will still be valuable regardless. The refactoring is not incorporated. It is more sensible to publish the fix first and then the refactoring later. It seems like a good candidate for a bigger overhaul.

Graveen:
:)
Can you see why appveyor is not correct ?
TY !

NetDwarf:
Looking into it for quite some time :D

NetDwarf:
Forgot to notify. It should now work.

Graveen:
;)

PR #259, opened 2020-10-06
Fix Linux Build

NetDwarf:


NetDwarf:
Should fix the separate folder issue on Linux where some files were stored in "debug" and some in "Debug".

Graveen:
Thx !

PR #264, opened 2020-11-15
Fixed placing and removing items in houses on 1125

Scrubtasticx:


NetDwarf:
The deletion of the requesting repo causes some problems with (git) diff of this commit. This would also affect git blame.

Graveen:
It is strange this is not a diff but complete files, nut i'll merge it btw.
Thank you for the work !

PR #266, opened 2020-11-28
Make (Stat)Buffs on NPCs more LiveLike 

NetDwarf:
Inlined behavior from PropertyChangingSpell for NPCs into corresponding Calculators. Caps were removed in the process and the different melee absorbs now multiply instead of add (e.g. `1 - ((1-x)*(1-y)*(1-z) ....)`). 
Magic "absorb" (which is in ResistCalculator) is still added.
Quickness behavior exists already for NPCs and WeaponSkill does not work on NPCs, so both behaviors are removed.

Note: On Live only BaseAF buff is used for melee absorb calculation, because the buff value is so high. I decided to use SpecAF as well for calculation, but set the cap to the BaseAF buff cap on Live (150 * 1.25 * 1.25).


NetDwarf:
Moved (most) behavior now further up into GameLiving.GetArmorAbsorb, so the Calculator now only manages the actual bonuses. 
This also removes (int) rounding errors.
GameKeepGuard lower limit of zero in GetArmorAbsorb is removed (and realistically never reached, but there is no reason to keep it).

Graveen:
Thank you !

PR #269, opened 2020-12-05
Add null check to DropLoot() for artifact scrolls

tegstewart:
When an artifact scroll couldn't be created, it lead to a null exception which stopped further loot templates from being processed and messed up mobs dying.

It now logs an error including the scroll data string so it can be added to item templates.


Graveen:
Nice mate !

PR #270, opened 2020-12-07
Adding Currency Exchange and Blood Seal Merchant

tegstewart:
Live no longer gives out glass and aurulite, instead granting BPs.  Live has also added a currency exchange mob that can exchange aurulite/blood seals/glass/scales.  See https://darkageofcamelot.com/content/1124-live-patch-notes for details.

I'm addressing both of those by adding ReceiveItem() to various merchants to allow special currencies to be exchanged for the type the merchant uses.  You can give scales to a blood seal merchant to get blood seals, or aurulite to a glass merchant to get glass.  You can also give aurulite and glass to BP merchants to get BPs, leaving them in the game but making them useful despite there not being any merchants that use them directly.

Both features can be disabled via server properties, and are enabled by default as they replicate live like behaviour.  I can see servers not wanting to let players grind aurulite or glass rather than participate in PvP, and some servers may want to keep aurulite/blood seals/glass/scales separate.

Blood seal merchants are present on Live but not in Eve yet, so I added it for future use.

I considered changing the loot generators for glass and aurulite to give BPs instead based on server properties, but that would require additional server properties and I think I prefer the items still being in game without them being completely useless.


Graveen:
Could we have a database table CurrencyExchange (id_nb source, id_nb dest, float rate) where we could fix the ratio ?
There are so magic numbers in the source code :)

But yes nice commit, as i said in https://github.com/Dawn-of-Light/DOLSharp/pull/271 i 'll btw commit it if you don't have time to improve it. 

Perhaps this table could also include the per-level scaling reward, ie the table used in dreadseals, but i'm shared with this; because it could lead to an overcomplex table, i don't see an easy solution

tegstewart:
I think this is an excellent candidate for a server property.

currency_converter_values: "AtlanteanGlass|1;Aurulite|4;DragonScale|40;BloodSeal|1000

Would that work for you?

Graveen:
Can you explain what is the SP number ? 1 glass = 4 aurulites = 40 scale =1000 seals ?

I have 3 suggestions:
- i think a table would be better, for both having multiples rates (ie, Glass -> X golds, Glass -> Y scales, Scales-> Z golds....)
source currency / target currency / conversion rate / min level for transaction 
- could you also ensure compatibility with Storm and/or inculde SQL for fixing this (ie content of table :D)
- could you see why it brokes Travis CI ? :)

Thank you !

tegstewart:
The error happens updating the compiler, and later PRs passed, so I suspect if we kick it hard enough to make it do a fresh check it'll pass.

It didn't occur to me at first how powerful a DB table could be when you first mentioned it.  That could be a truly general case solution in which a CurrencyConverter class could convert anything to anything else.  It could even handle exchanging champion weapons by allowing conversions from champion weapons to champion weapon tokens.

This works as an interim measure, but nobody has asked for it, and I don't really need it myself.  I think we can put it on the back burner for now and I'll look into a table based currency converter down the road.

NetDwarf:
The solution is probably way simpler in mechanics (not implementation!). On Live the currency exchange merchants are just handled like any other merchant selling items (e.g. ingredient merchants where you can buy stacks). What we miss is the currency (regardless of exchange implementation).

tegstewart:
I'm going to close this for now.  We don't need an interim implementation, and it'll be a while before I have time to do this more elegantly.

Graveen:
I prefer put it "as is" even if you don't have time to improve. This is in the mood of the code you are upgrading, not better, not worse. Please reopen it you still accept the merge. My comments are here to point the way we could have a better implementation 👍

tegstewart:
I suppose it doesn't hurt anything and doesn't add much work if I come back to do a more elegant implementation.

The more I think about it, the more powerful a table seems.  For the following columns:
ClassType, MinLevel, InItem, OutItem, Ratio
With checks for xp, bounty points, and realm points in the OutItem entry, the code could go into GameNPC.ReceiveItem() and handle everything from currency conversions to dreaded seal collectors to xp granting token items, all controlled by the DB.  *Head explodes*

NetDwarf:
On Live you can have different currencies on one merchant however I can't remember any merchant that had different currencies one page.

Idea: Have a special and (probably negative) slotposition for a (merchant) page that can signify a different currency which is an item_template (id_nb). When you buy instead of deducing copper, item_template*count is deducted (similar to how crafting works). By default (if this slotposition is missing) the prices are in copper (backwards compatible).

Currency exchange would be done via a merchant that sells a currency item_template for another currency item_template. No conversion table necessary :)

tegstewart:
That seems like a much more versatile solution that the current scads of hardcoded special currency id_nbs for merchant classes.

It could also allow for multiple currencies on one merchant, by page or even by item.  Maybe a given merchant sells the same items on four different pages, with the page choosing the currency being used.

Graveen:
Merging as is, before the .Net upgrade

PR #271, opened 2020-12-07
Fix dreaded seal collector

tegstewart:
More valuable seals created via crafting are given their proper values rather than all seals being treated equally and giving somebody turning in an effulgent dreaded seal 1/250th of it's value.

The collector now takes seals and grants BPs when the player is at max realm rank rather than refusing to take the seal entirely.

There was a fair bit of clean up as well.  The script this was based on had several people fiddle with it before it got to me, and there was a lot of room for improvement.  It's possible I'm better at this than I was three years ago.


Graveen:
The script is better, no doubt :)
I understand the problem here is the scale-per-level that us hard to make database driven.

Should be reworked to integrate all this in database btw, because when i read the switch based on id_nb and static values, omg :) 
At least your work is gathering all theses magic values in a few arrays and this is more easily readable/updatable.

Could you add a Dictionnary<MaxLevel, Rps> for better integration ? Thus it would really put static datas out of the code, and replace the long trail if (Level < x) amount = z ....?

Anyhow if you can't i'll accept this PR of course :)
Thank you VM !

tegstewart:
The original script was determining RPs by dividing the price of seals.  It didn't work when the DB price values were all the same, but I fixed that last night, so we can go back to it.

A server property could be added which includes the levels and multipliers as a string akin to how mob styles are handled.  I'm hesitant to do that, mostly because I'm trying to curb my adding lots of server properties addiction.  I suppose it's just one more server property, not really a big deal.  *Slaps forearm and readies needle*

Would that work for you?

Graveen:
Perhaps the currency table discussed in the other PR.
For now a simple SP or dictionnary in code is ok.

PR #273, opened 2020-12-08
Random loot generator fixes

tegstewart:
When zone realm wasn't specified, no item template would be selected.  Since most DB releases don't assign realms to zones, it never works.  If the zone's realm is not valid, a random player realm is now chosen instead.

Only items that can be equipped are now added to loot templates, to stop things like trophy remains, saddles, and rent token from dropping.

Changed to use a static constructor rather than keeping a superfluous copy of all item templates solely to see if loot templates have been populated.

Switched from AddRandom() to AddFixed(), as the former causes mobs dropping trash to block the item drop.

The hard coded 10% drop chance could be changed to a server property, but I think it might be a better idea to add a drop chance column to the LootGenerator table to define how often items drop for all loot generators rather than adding a server property specifically for this one.


Graveen:
@NetDwarf  what do you think about this PR ? :)

NetDwarf:
Okay, what I noticed: 

1. The (preexisting) `if (itemTemplates != null)` is unnecessary if I'm not mistaken (SelectObjects should never return null)
2. There is a lock missing while the arrays are populated (access to default/null value possible which is hard to debug later)
3. static constructor hides what it is doing (and make a comment necessary, that otherwise isn't)[[1]](https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#names-rules) (extract method would do)

Small tip: to avoid "Syncing to base" you can just do `git fetch upstream && git rebase upstream/master` where upstream is the name for the DOLSharp repo [see this if not set](https://git-scm.com/docs/git-remote)

tegstewart:
The lock made sense when the standard constructor was populating the arrays, so you could have multiple instantiations and the first might still be populating the arrays when the next see the arrays exist and tries to access them.  Shouldn't a static constructor instead force the arrays to finish populating before any run non-static methods can be called, making the lock redundant?

Once my other PRs are resolved, I'll do a fresh fork to get rid of the "syncing to base" nonsense.

NetDwarf:
> Once my other PRs are resolved, I'll do a fresh fork to get rid of the "syncing to base" nonsense.

Not necessary here, really. It was really just a small tip you might use in the future .. or not :)

> The lock made sense when the standard constructor was populating the arrays, so you could have multiple instantiations and the first might still be populating the arrays when the next see the arrays exist and tries to access them.

You're right, (I) strike that :grin:

NetDwarf:
:+1:

PR #274, opened 2020-12-10
Fix keep healers

tegstewart:
The guard's target was changed to the target to be healed before the LOS check was sent, but would change to a hostile target before the LOS handler was called.  As a result, once the LOS handler was called, the target would now be hostile and healers wouldn't ever actually heal.

Added a HealTarget property to independently store the target to be healed, so it doesn't matter if the healer switches to a hostile target before the handler is called.  Healers work properly now.

Also assigned players as LOS checkers in the first GetPlayersInRadius() call to hopefully save an unnecessary GetPlayersInRadius() call later.


tegstewart:
It looks like Travis threw a horseshoe when trying to update fsharp.

Graveen:
Nice TY ! Can you see why it breaks Travis plz ? 👍

PR #275, opened 2020-12-23
Update to .Net Framework 4.8 and all packages

NetDwarf:


NetDwarf:
This needs some review also by me.

SharpZipLib had some API changes which are relatively safe to refactor to, since it doesn't work in the current state anyway (wrong Endianness in headers). I will submit a fix later, which is already done, but it's no priority, since it is unused anyway.

MySql.Data to 8.0.22 introduces new dependencies which make it necessary to include netstandard.dll for script compilation, but only on Windows. A step back would be an update to 6.10.9, which does not need that workaround. 

The conversion to new csproj involves a lot more changes and uses new defaults and has different environment variables, so this might introduce problems. Especially the default that every file is now included added new broken scripts which needed patching (safe changes in themselves). 

All this can be used as a stepping stone to .Net Core/.Net5+ conversion, because these changes are recommended anyway.

Graveen:
Did you check versus Storm ? :) I'll merge and we will see. Yes we shoudl go to .NET Core, and the sooner the better imho.

NetDwarf:
Oh that was quick. I actually wanted to make some checks and maybe someone look into the changes made, if one finds any errors. Phew. That is a risky one :smile:

Graveen:
I think in any cases there would be some related patches 👍 so let's take that road !

Graveen:
and we could revert quickly hu hu :D

PR #276, opened 2020-12-24
Fix MPK utility

NetDwarf:


NetDwarf:
This fix was a happy little accident while updating actually :smile:

Graveen:
I'm not understanding the missing 3 bytes, but i guess it is totally normal :D
Thank you !

NetDwarf:
The worst case was with 4bytes where only 1byte was actually preserved, so I offsetted the buffer by 3 bytes :grin: That was pretty annoying to find actually.

PR #277, opened 2020-12-25
Downgrade MySql.Data to 6.10.9

NetDwarf:
The compatibility layer in version 8 introduces a big performance impact


NetDwarf:
The usage of netstandard.dll will likely be necessary again when actually upgrading to .Net5. I hope MySql.Data 8 is more performant then.

NetDwarf:
Travis should be fixed with #276

PR #278, opened 2021-01-07
Revert ControlledNpcBrain.Get..Owner() methods

NetDwarf:


NetDwarf:
The bug broke any charmed pets.

PR #280, opened 2021-02-02
Origins

Shursan:
Added code attack on Gameliving for be able to attack another target that the main target if this one is out of range. 
Exemple with Pet: if the owner take aggro and become the main target, the mob will try to hit anyone in his aggrolist that is in attack range before he can reach the owner of the pet..

Added use item desactivation for 2 mn (i don't know what's the timer on live) on new equipped items.

### DIFF
--- a/DOLDatabase/Tables/Inventory.cs
+++ b/DOLDatabase/Tables/Inventory.cs
@@ -443,7 +443,14 @@ namespace DOL.Database
 
 		public virtual int Weight
 		{
-			get { return Template.Weight * m_count; }
+			get
+			{
+				if (Template.PackSize > 1)
+				{
+					return (int)((double)Template.Weight / (double)Template.PackSize * (double)m_count);
+				}
+				return Template.Weight * m_count;
+			}
 			set { Template.Weight = value; }
 		}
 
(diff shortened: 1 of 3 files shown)


NetDwarf:
Thanks a lot for your PR. 

The attack behavior for mobs seems to work fine and is actually something that was really missing. Thumbs up for that! :+1: 

The timer reset on reequip of item (to 2 min) did not work in my testing, however reverting to the previous commit did work.

This is just a headsup. I am going to merge this anyway in a few days since it doesn't seem to break anything. If you're not fixing the second aspect of the PR now you can fix it in a separate PR :smiley:. In general having separate branches for non-tangential fixes is a good way to separate concerns.

Shursan:
Gotcha for the separates branch, thanks for advice :)

Shursan:
@NetDwarf are you sure you have tested the timer reset on equip as a player ? there's no timer reset apply as a GameMaster

NetDwarf:
Yeah, I am sure. It also worked when I reverted the last commit.

Shursan:
With that we also have a TempProperties in Memory save.. that can register TempProperties at player disconnect and reapply on player connecting again... it was configurable by serverproperties, serialized string for choose only what can go in. I will look soon if ti can be interesting for DOL

PR #283, opened 2021-02-09
Temp properties manager

Shursan:
TempProperties registering possibility.
Save in Memory and can be applied again when player entered Game.
You can configure what can go in using  serverproperties.


NetDwarf:
Let's talk on Discord! I sent you a PM :smile_cat:

PR #284, opened 2021-02-09
Introduce PlayerRace class

NetDwarf:
Inline TemplateMgr and SpellMgr
Refactor KeepGuard


NetDwarf:
This is supposed to be a refactoring overall with only one actual change to behavior, which is that guards can have slightly different races now. I did that because it was more natural to use a CharacterClass as a base for the selection of races for the semi class based keep guards (and Doppelganger).

PR #286, opened 2021-02-10
TempProp Manager

Shursan:
Allowing tempproperties to be registered and reapplied on player connect


NetDwarf:
Okay, just tried and seems to work like a charm. :+1:

PR #287, opened 2021-02-14
Crafting memory possibility

Shursan:
Added crafting memory possibility.
Added crafting sell price control to livelike (95% of cost)
Added crafting memory reload command (can be used for clean up memory on change)


Shursan:
maybe for works fine it need tolower id_nb into database
i don't want to cast tolower check in c# as it taking much time.

UPDATE itemtemplate
SET 
    id_nb = LOWER(id_nb);

Shursan:
sellingpricecontrol

NetDwarf (changes requested):
As always review is done in good faith and not meant as an offense :sweat_smile: So without further ado here are my requests:

1. Encapsulate members of CraftedItemCache
2. Move caching mechanism out of GamePlayer (not its responsibility)
3. in addition also separate behavior (added code in GamePlayer.CraftItem does at least two different things)
4. Remove commented code (or uncomment, but I didn't look into it)
5. Is a new Serverproperty really necessary? Any contraindication to just force it?
6. Require the least amount of changes to DB as possible (since it's a caching mechanism it should require no changes to DB)

Shursan:
6/ As i remember it was done like this because:

SQL checks find results if you write all in uppercase or lowercase.. 
but into memory list it is not the same, if u got a record into crafteditem with id_nb = TEST_ITEM and the itemtemplate is Test_Item, it will not find it.. It's why it's required.

5/ I always prefer to give choice to use it or not.

4/Maybe all the commented code can be removed... It's just if u want to load all into memory at startup. first time i have write it i have done it like this, but after somes reflexions, we have choosen to only loads into memory crafteditem on runtime when used in game.

Graveen:
Nice to see your PRs here @Shursan 

On my side, i would like to better understand why you implemented this: what is the initial issue, and what are the drawbacks. Did you notice a bottleneck on DB accesses ? What is the final size of the cache ? Any idea of the .ToLower() overhead (could be long but is done once) ? Last but not last, i see 95% of the price. This 95 would deserve an SP.

Based on the answer the SPs could be totally inutiles - if we are under reasonnable memory consumption and cache building time, the cache way must be used.

Last thing, i think we should rename the english text (and the SP too) - it is confusing. IE 'Reset crafting memory' could be 'Craft RAM cache reload"

This is an interesting PR to discuss, because it somehow twisting things. The database role is to serve datas, while gamserver.dll is a client. While implementing caches, we are twisting roles and perhaps improving DB layer instead would be far better. But it is a big big topic !

Thank you for the work @Shursan  :)

NetDwarf:
Sorry for answering so late.

I'm going to make a test for GamePlayer.CraftItem so we can measure the actual effect. However we also need to do something to improve readability and therefore maintainability of this patch.

NetDwarf:
Okay, I tested a lot of stuff, tho I'm going to commit only one explicit test.

My findings: 
- CraftedItemCache/-Manager is about 2-times faster than native implementation
- I tested both the newest and the first commit only and found no performance difference (which is to be expected, but I wanted to know)
- test with a recipe where the ingredient items are missing was exactly as fast the working recipe with CraftedItemCache (I removed that test, because there is no point in testing broken recipes)
- database tables have a PreCache setting which - when enabled on CraftedItem CraftedXItem - consumes  roughly 12MB of additional memory with a full crafting DB
- with PreCache enabled the speed is about the same to CraftedItemCache/-Manager

I was not aware of that setting until I made that test. 

My proposition is therefore #288 instead and add the price correction mechanism afterwards.

NetDwarf:
#289 should address performance and price change. I'm therefore closing this PR.

PR #289, opened 2021-02-20
Add option to adjust Recipe.Product.Price

NetDwarf:
Add Recipe class
Add RecipeDB which caches Recipes on access
Add Recipe.SetRecommendedProductPriceInDB (implementation from @Shursan )
Add ServerProperty to control the setting of recipe product prices (opt-in)


NetDwarf:
Notes: 
- performance test (1000x call to Player.CraftItem on existing recipe) got down from ~4s to 40-70ms
- RecipeDB also caches invalid recipe queries to avoid ObjectDB queries

I mostly tried to keep behavior as is, however I removed one segment:
`string adjItem = itemToCraft.Id_nb+(GetQuality(player, recipe).ToString());`
`ItemTemplate adjItemToCraft = GameServer.Database.FindObjectByKey<ItemTemplate>(adjItem);`
`if (adjItemToCraft != null)`
`{`
`        newItem = GameInventoryItem.Create(adjItemToCraft);`
`}`
`else`
`{`
`        newItem = GameInventoryItem.Create(itemToCraft);`
`} `
and replaced it with this one:
`newItem = GameInventoryItem.Create(product);`

It was done in an attempt to simulate quality based consumables and such, but was never actually done in DB - at least not on Storm. This should be done another way if at all (it serves no purpose on Live afaik).

Graveen:
Very nice TY !

PR #290, opened 2021-03-05
Spell handler for the style effect 'power rend' intended for use with the Valkyries Sword spec line styles Aurora (15) and Odin's Bite (40).

Xanxicar:
* Moved class PowerRendSpellHandler out of Convoker.cs where it is being abused to poorly simulate a power trap spell.
 * Added PowerRendSpellHandler.cs as a separate code file declaring it a Valkyrie only melee style effect.
 * Changed logic to reflect Pendragon test results.
 * Changed messages to Caster and targets.
 * Improved precision of the amount of power drained.
 * Added random variance.
 * Added spell casting interruption.
 * Added spell resist chance.
 * Added breaking of mesmerization, root and snare.

### DIFF
--- a/GameServer/spells/Masterlevel/Convoker.cs
+++ b/GameServer/spells/Masterlevel/Convoker.cs
@@ -748,47 +748,6 @@ namespace DOL.GS.Spells
 		}
 	}
 	#endregion
-
-
-	#region PowerRend
-	[SpellHandlerAttribute("PowerRend")]
-	public class PowerRendSpellHandler : SpellHandler
-	{
-		public PowerRendSpellHandler(GameLiving caster, Spell spell, SpellLine line) : base(caster, spell, line) { }
-
-
-
-		public override void FinishSpellCast(GameLiving target)
-		{
-			m_caster.Mana -= PowerCost(target);
-			base.FinishSpellCast(target);
-		}
-
-		public override void OnDirectEffect(GameLiving target, double effectiveness)
-		{
-			if (target == null) return;
-			if (!target.IsAlive || target.ObjectState != GameLiving.eObjectState.Active) return;
-			//spell damage shood be 50-100 (thats the ammmount power tapped on use)
-			int mana = (int)(Spell.Damage);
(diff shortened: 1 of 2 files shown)


NetDwarf:
Thank you! Everything seems to work as intended. :+1: 

I'm going to merge it in one or two days anyway however you might want to remove the comments. Your naming for variables and methods is pretty much spot-on and so your comments become redundant, which - again- means the naming is good, but also that you can remove the comments. Conversations also don't belong in code (but here). I'd say none of the comments are necessary (except the license of course). 

The gist of how to comment code is pretty much: Only do it if absolutely necessary and be concise, but for a more detailed explanation read [this](https://mosquitolwz.medium.com/clean-code-quotes-4-comments-2de28ba0a6e2)!

Xanxicar:
Thank you for your feedback @NetDwarf ,
I have edited the initial post to include much of the information from the comments in the source file so that I can remove it from there. When I wrote the file a year ago I wasn't sure about the target audience and got a little verbose. A good chunk of it was actually meant to go here but it had been so long that I forgot.

What is your stance on XML doc comments ? The rest is already on its way out.

Graveen:
Nothing more to say, the commit's description is really nice !
Thank you !

NetDwarf:
Whoa, very nice and detailed description. :+1:  

> We found absolutely no way of scaling the effect.

My guess is that it plain works like % cost. That means that the rend just uses *base* mana of the victim. Of course in contrast to % cost it is unaffected by skill, focus etc. That being said, we just keep the current implementation regardless, if there is no new insight.

> What is your stance on XML doc comments ? The rest is already on its way out.

Only for external interfaces basically. Because DOL does not really have that I'd say it's okay for interfaces in general. Most of the time these comments just repeat what the method name (including the class name) is already telling. Which is pretty much the case here.

Xanxicar:
@NetDwarf all cleaned up and good to go.

Graveen:
@Xanxicar can you please share the SQL here ? I could apply to Storm ;)
Thank you !

Xanxicar:
@Graveen added a zip file with .sql and .csv inside under the USAGE section for convenience.

PR #292, opened 2021-03-13
Fix Banners not loading

NetDwarf:
Fix some guards not loading when a component with same Skin and Rotation exists multiple times in a keep


NetDwarf:
**Please do not merge before the "FixKeeps" Storm commit**

Side note: Banners on Towers do not appear on first start, but when captured.

Graveen:
Ok merged both :)

PR #293, opened 2021-03-13
Fix ObjectDatabase.SelectObject

NetDwarf:
Correctly return null when no item is found


NetDwarf:
There was a mismatch when using parameterized version vs unparameterized version. The latter returned null before, but the former errored out. Since finding no item is a pretty normal result an Exception is not appropriate behavior.

PR #294, opened 2021-04-02
WhereExpression as (single) parameter object

NetDwarf:


NetDwarf:
This is a simple abstraction from the verbose parameterized queries, so should still do the same input sanitation as before. The QueryParameter placeholders are generated.

PR #295, opened 2021-04-02
Bug Fix: Mobs adding through floors/ceilings/coming out of walls

Xanxicar:
This small change to StandardMobBrain.CheckPlayerAggro() ensures that the Z-axis is respected in dungeons analogue to how it is done in CheckNpcAggro().

Mobs in dungeons will no longer aggro on players in rooms far above or below them.
In Darkness falls in particular this would cause body pulls of high level mobs appearing seemingly out of thin air or coming out of walls.


NetDwarf:
Thank you! :+1:

PR #296, opened 2021-05-22
Many fixes (& small improvements)

JVirant:
fix remove virtual for helper methods
fix moc calculation
small optimization about Regions and InRange functions (use of struct instead of class, less GC usage)
change cmd info: add more infos + formatting
fix nullref about object spec level + better customization of server rules
add new version support of log4net+dotnet core
fix player loaded event and kick everyone in case of exit
add dps, spd, af, abs to npc (not necessarily used)
change NPC's pathing: when a npc return to the nearest point instead of first one
change item info: always show usable classes and description
fix npctemplate loading for charisma and a nullref
misc fixes:  nullref, error with some xml config, fix /keep create and remove
fix overloading of GameLiving.MakeAttack
fix duplication of saved player effects
change log4net_unix: use AnsiColorTerminal instead of ColoredConsole
fix respec command, clear properties when declined
fix command /account movecharacter
change standard brain will buff friendly npc in range
fix typo about acuity buff in StandardMobBrain
change when a npc return to the nearest point instead of first one
fix controlled npc brains' GetLivingOwner/GetPlayerOwner/GetNPCOwner
fix StandardMobBrain.AddToAggroList overrides
fix animist turret: face its target and SpeedDecrease spell
activate VT100 emulation for windows 10 (useful to keep the same log4net config between windows and unix)


Graveen:
Tons of fixes, very nice ! Thank you

PR #297, opened 2021-05-23
Miscfixes

JVirant:
fix remove virtual for helper methods
fix moc calculation
small optimization about Regions and InRange functions (use of struct instead of class, less GC usage)
change cmd info: add more infos + formatting
fix nullref about object spec level + better customization of server rules
add new version support of log4net+dotnet core
fix player loaded event and kick everyone in case of exit
add dps, spd, af, abs to npc (not necessarily used)
change NPC's pathing: when a npc return to the nearest point instead of first one
change item info: always show usable classes and description
fix npctemplate loading for charisma and a nullref
misc fixes: nullref, error with some xml config, fix /keep create and remove
fix overloading of GameLiving.MakeAttack
fix duplication of saved player effects
change log4net_unix: use AnsiColorTerminal instead of ColoredConsole
fix respec command, clear properties when declined
fix command /account movecharacter
change standard brain will buff friendly npc in range
fix typo about acuity buff in StandardMobBrain
change when a npc return to the nearest point instead of first one
fix controlled npc brains' GetLivingOwner/GetPlayerOwner/GetNPCOwner
fix StandardMobBrain.AddToAggroList overrides
fix animist turret: face its target and SpeedDecrease spell
activate VT100 emulation for windows 10 (useful to keep the same log4net config between windows and unix)


NetDwarf:
Thank you for your work in advance. I'm going to review every commit individually. 

One thing I want to say upfront is, that generally speaking it is best to split up PRs to the concerns addressed, this is especially true if the concerns are orthogonal to each other (i.e. logging fixes and NPC absorb). 

Mixing concerns also leads to potentially fixing over multiple concerns at once which muddies the waters further. So it's good practice to keep PRs concise. In this case you already splitted up your fixes nicely with commits, so it's probably easy to split it up again. 

At any rate. Keep up the good work! :)

JVirant:
Sorry for that, it was just many things done on Amtenael which wasn't splitted from custom behaviors to be in DOL

NetDwarf:
No problem, it just needs splitting up afterwards, but please wait for the review before you do that. 

It's generally good practice to make a new branch for each concern (and then merge them to your production branch). This has multiple advantages. The most obvious is, that you can share the branches directly as PR, but it is also advantage while developing to work only on one concern and not multiple.

JVirant:
I'm working on the pathing system on top of this branch, I will not split it.

NetDwarf (commented):
tl;dr: PR can't be approved as is. The main reason is that it breaks Storm (and likely every other server using DOL). Additionally there are explanations missing for a lot of changes.

# Fix log4net output across terminals
* e39030b53 activate VT100 emulation for windows 10
* e77a54b6c change log4net: use AnsiColorTerminal instead of ColoredConsole
* 1bd852d58 add new version support of log4net+dotnet core

This can be fixed by using ManagedColoredConsoleAppender (and removing Highlight in config)

# Fix turrets not facing correctly
* f47169da9 Fix turret: face its target and SpeedDecrease spell

Why change the think interval? Would it be possible to remove facing requirement for turret (or even NPCs in general) instead?

# Fix NPC AI (Brain)
* e5b0cdf5f fix StandardMobBrain.AddToAggroList override

What is it specifically fixing?
Do not remove virtual from methods that might be used by scripts (Basically: Don't remove virtual at all)
* 8cdb36621 fix controlled npc brains' Get**Owner

What is it fixing?
Do not remove virtual from methods.
* 88e3b9f6c change when a npc return to the nearest point instead of first one
* b1987a5eb change when a npc return to the nearest point instead of first one

What's the reason for change?
* d36e1411a fix typo about acuity buff in StandardMobBrain
* 251394144 change standard brain will buff friendly npc in range

Does not seem to work (Tested on Storm/Arawnite Shamaness)

# Fix command /account movecharacter
* b6f1ac7c9 fix command /account movecharacter

Works like a charm.

# Fix /respec
* 258b30d2f fix respec command, clear properties when declined

What problem did it cause before?

# Fix duplication of player effects
* d0a164fed fix duplication of saved player effects

What triggers duplication?

# Change signature of GameLiving.MakeAttack
* cd27474ea fix overloading of GameLiving.MakeAttack

Changing signature breaks every server overriding GameLiving.MakeAttack in scripts

# Fix NPCTemplate loading for Charisma and adding null guard
* b5cdff2e4 fix npctemplate loading for charisma and a nullref

What can trigger FriendFaction being null? It would be better to avoid it being null than checking for null.

# Add DPS, SPD, AF and ABS to NPC
* 7cdce875a add dps, spd, af, abs to npc
* 067bbb1cc fix GameNPC.GetArmorAbsorb with last changes
DPS, AF, Abs, Spd are redundant in NPCTemplate database. They are calculated by Str, Con, Qui.

# Make item delve more verbose
* 36ef9ceac change item info: always show usable classes and description

# Fix player loaded event and kick everyone when server is stopped
* aed059bad fix player loaded event and kick everyone in case of exit

What problem was this causing before?

# Remove HealSpellHandler.HealTarget(GameLiving, int)
* 2a65070b0 fix remove virtual for helpder methods

This breaks every server using it.

# Show more information on /mob info and /GMinfo
* 86d8bd024 change cmd info: add more infos + formatting

NPCs do not rely on Endurance, Mana, Conc so the information seems superfluous at least for /mob info

# Fix AbstractServerRules.GetCompatibleObjectTypes
* 8f26b7c3b fix nullref about object spec level + better customisation of server rules

Is null as SpecName expected? If not, it should trigger an error.
Changing GetCompatibleObjectTypes to be public breaks all servers overriding this.

# Add ctor for ZoneData and change PlayerDistEntry+NPCDistEntry to struct
* 983eed817 small optimisation about Regions and InRange functions

What does it optimize?

# Change MOC calculation
* d467e8d53 fix moc calculation

What was wrong with it?

# Misc Fixes
* c29020958 misc fixes

This needs to be split up and/or a better explanation

NetDwarf:
Regarding splitting up branch: This is usually done by using (different) fresh branches and cherrypicking commits over from this one (with interactive rebase or cherrypicks for example). I know it's overhead, but important when collaborating.

JVirant:
> There is no denying that breaking changes can be necessary, but they need a good reason. I can't find a single reason to remove `virtual`. Most of the time you can get away with obsoleting methods instead of removing them without replacement. Refactoring old methods to use the new one (or vice versa depending on the situation) and obsoleting the old method is an option, too.

By removing the virtual (and not the method), it create an error for something that is probably already broken in a script but visible only on edge cases.

> There is always a good chance that the code endeavour ends up not usable or you need to take some steps back to still keep it compatible. That compatibility is necessary if multiple parties use your code. If you break stuff on purpose it should better be worth it (and that is indeed a possibility).

Ok, I don't have time for that and my next refactor change a very low level part of DOL so I continue on my branch, I close the PR since no one will continue to work on it...

PR #299, opened 2021-05-29
Fix colored logging text on most terminals

NetDwarf:
Replace ColoredConsoleAppender with ManagedColoredConsoleAppender
Remove unix specific logconfig.xml
Fix colored console output for most (if not all) terminals


NetDwarf:
Tested it on cmd-terminal, powershell-terminal and gnome-terminal.

Graveen:
Thank you !

PR #300, opened 2021-05-29
Deprecate explicitly parameterized SelectObject(s) queries

NetDwarf:
Use implicit version instead
Rename WhereExpression to WhereClause
Replace all SelectObject(s) queries with implicit parameters (WhereClause object)
Replace many internal calls in *ObjectDatabase with implicit parameters

### DIFF
--- a/DOLDatabase/WhereClause.cs
+++ b/DOLDatabase/WhereClause.cs
@@ -22,7 +22,7 @@ using System.Linq;
 
 namespace DOL.Database
 {
-    public abstract class WhereExpression
+    public abstract class WhereClause
     {
         private static string alphabet = "abcdefghijklmnopqrstuvwxyz";
 
@@ -40,7 +40,7 @@ namespace DOL.Database
             return prefix + result;
         }
 
-        public virtual string WhereClause
+        public virtual string ParameterizedText
         {
             get
             {
@@ -63,7 +63,7 @@ namespace DOL.Database
             }
         }
 
-        public virtual QueryParameter[] QueryParameters
+        public virtual QueryParameter[] Parameters
         {
             get
             {
@@ -83,12 +83,12 @@ namespace DOL.Database
 
         internal abstract List<TextAtom> IntermediateRepresentation { get; }
 
-        public virtual WhereExpression And(WhereExpression rightExpression)
+        public virtual WhereClause And(WhereClause rightExpression)
             => rightExpression.Equals(Empty) ? this : new ChainingExpression(this, "AND", rightExpression);
-        public virtual WhereExpression Or(WhereExpression rightExpression) 
+        public virtual WhereClause Or(WhereClause rightExpression) 
             => rightExpression.Equals(Empty) ? this : new ChainingExpression(this, "OR", rightExpression);
 
-        public static WhereExpression Empty => new EmptyWhereExpression();
+        public static WhereClause Empty => new EmptyWhereClause();
 
         public override int GetHashCode() => base.GetHashCode();
     }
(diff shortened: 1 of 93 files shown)


Graveen:
nice refactoring / code quality improvement ! merge when you go out of WIP

NetDwarf:
Yeah, need to fix merge conflicts. Then it should be ready. :smiley:

Lotendan:
Hello @NetDwarf, have you had the chance to look at the performance implications on the db/server of the refactoring?

NetDwarf:
At least in the limited testing you can do locally, it performs about the same. This includes Integration tests (one of them performs 100 queries at once via MultipleSelectObjects). The other one is starting a custom server locally and stopping time. 

I did not yet test the very last commit with the server, but it uses the "old" parameterized queries under the hood. The only overhead is the query translation itself. This can actually also be removed at some point, but it won't help with performance I think.

JVirant:
> Hello @NetDwarf, have you had the chance to look at the performance implications on the db/server of the refactoring?

Against the old implementation, it's way worse :
- Allocating at least 4 more objects (List + 3 TextAtom) which take pressure on the GC
- Many indirections from theses objects : TextAtom (a pointer) contains an object (another pointer)
- virtual used in many parts: it add another pointer indirection
- a loop to create the parameter's id is bad especially with the IN operator which can contains many parameters

But even with theses changes, it's mostly not visible because executing the sql query itself is way longer than that. The important thing, it's the GC pressure, creating too much object can be a real issue for a game.

If you need to squeeze the maximum performance, use the method `IList<TObject> SelectObjects<TObject>(string whereExpression, IEnumerable<QueryParameter> parameter)` perhaps the one without queryparameter is faster but you can have some security issues...

NetDwarf:
> But even with theses changes, it's mostly not visible because executing the sql query itself is way longer than that. The important thing, it's the GC pressure, creating too much object can be a real issue for a game.

This is a non-issue really. I test on the worst config possible (i5-8250U Windows-VM on Linux/Qemu) and it still does not make a difference. Do not fear to create objects.

Lotendan:
I tend to agree with @JVirant about GC pressure. Sure it doesn't make a difference in our local test bed, but this might become an issue under workload.
On the other hand I like the changes in this PR since they bring an abstraction around the db. However, for the few paths that are performance-sensitive, I would maybe not deprecate the "old" APIs with the QueryParameters (not talking about the one without QueryParameter, this one definitely needs deprecation due to security issues it may cause) and leave the developer with options as to whichever is better with or without abstraction in the specific context.
What do you guys think?

NetDwarf:
> I tend to agree with @JVirant about GC pressure. Sure it doesn't make a difference in our local test bed, but this might become an issue under workload.

It's irrelevant. If you aren't pumping out 2 million (simple) objects a second you optimize the wrong thing. If there is no functional issue found with this PR, this is going to be merged :stuck_out_tongue: 

As a side note: That's also why value objects are a thing. And I am going to use them when I can.

DOLDatabase/WhereClause.cs
```diff
@@ -0,0 +1,265 @@
...
+            do
+            {
+                var letter = alphabet[(int)(id % radix)];
+                result = letter + result;
+                id /= radix;
+            } while (id > 0);
+            return prefix + result;
+        }
+
+        public virtual string ParameterizedText
```
JVirant, line 43: It should not be here, the SQL syntax can be different between 2 dbms so the IR to SQL should be done in the SQL implementation.

I think it's the whole purpose of using WhereClause instead of sql string directly if it can't do that, all theses changes are way less interesting.

JVirant (changes requested):

DOLDatabase/WhereClause.cs
NetDwarf: The purpose of this is just to simplify the old parameterized queries. If you someday decide to use a completely different (Non-SQL) DBMS you can still refactor it, but otherwise YAGNI applies.

JVirant: `Type` is a reserved keyword and it's used at least for the Ban table so it's needed.

NetDwarf: What is the scenario you're testing there?

JVirant: I didn't tested it, I'm just saying that this PR is adding complexity (and possible bugs with it) for nearly nothing if it can't manage simple scenario like having a sql keyword as a column name.

If I follow the YAGNI principle, this whole PR is against it: it add complexity for doing less. When you write the sql where clause directly, you can do some subqueries, use sql functions, etc. And you don't need to learn a new way to query the database.

NetDwarf: As I said above the purpose of the original change was to simplify them and that includes easing the transition to parameterized queries. The process of writing them was error prone and repetitive with no IDE help. In addition the WhereClause queries are also easier to read.

JVirant: Ok, I see. Sorry for losing your time about that!

PR #301, opened 2021-05-29
fix DataQuest, give the same item 2 times in some cases

JVirant:
Fix a small bug in DataQuests: if the quest has an item to give at step N (seems fine) but if the next step N+1 is a deliver step, it will give the next item in advance (seems wrong, when you finish deliver step, you get the item once more).

I don't know why this code was added in the first place, I think I removed the wrong part.

### DIFF
--- a/GameServer/quests/QuestsMgr/DataQuest.cs
+++ b/GameServer/quests/QuestsMgr/DataQuest.cs
@@ -1642,16 +1642,6 @@ namespace DOL.GS.Quests
                         stepTemplates[0] = template[0];
                     }
 
-					if (nextStepType == eStepType.Deliver || nextStepType == eStepType.DeliverFinish)
-					{
-						// Allow StepItemTemplate to be empty, assume quest player received item in a previous step or outside of the quest
-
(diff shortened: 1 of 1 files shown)


JVirant:
Actually, I found the issue, my PR is wrong:
```
	/// StepItemTemplates - Any items that need to be given to the player for a step.  Every step can give an item to a player. All
	/// steps give an item at the completion of the step except Delivery and DeliveryFinish.  If StepItemTemplates are defined for a 
	/// Delivery step then the item is given at the beginning of the step and accepted by a target to end the step.
	/// For Kill and Search steps, StepItemTemplates can include a drop chance behind the template name.  Ex: |some_template_name;50|  
	/// If the item does not drop then the step is not advanced.
	/// If no items are given to a player at any of the steps then this can be null, otherwise it must have values for each step. 
	/// Empty values || are ok. 
```
It's a special case for Delivery steps, I think we should make Delivery steps deprecated and use Collect which seems more intuitive.

In all case, there is a bug but I think the solution is to add a if in the generic part giving the StepItemTemplate.

Graveen:
@JVirant ah ok ! i close this PR if it fixes nothing :)

PR #302, opened 2021-05-30
Add -crashonfail option to ConsoleStart

NetDwarf:


NetDwarf:
This is a step for making it harder for "YaY to break Storm". :smile: 

If server terribly fails with `-crashonfail` option set, it exits with non-zero value. This in turn makes it possible for scripts using DOLServer to counteract this (for example by returning to the known good state).

PR #305, opened 2021-06-02
Whereexp isin

JVirant:
Add `DB.Column.IsIn` operator which transform to sql `IN`, it's a bit more complex than other operators but it can save many SQL queries if it's used correctly. There are 2 usages of this optimization, we can find many more in the code but it's not obvious.

Also, I added a small unit test for `FindObjectsByKey`.

### DIFF
--- a/DOLDatabase/WhereExpression.cs
+++ b/DOLDatabase/WhereExpression.cs
@@ -18,6 +18,7 @@
  */
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace DOL.Database
 {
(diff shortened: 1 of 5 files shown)


NetDwarf:
I did not expect adoption so quickly, but this is actually a better way to do "multi"-queries and I wonder if we could just replace them with SQL primitives. 

Please note that `WhereExpression` is going to be renamed to `WhereClause` and the properties `.WhereClause` and `.QueryParameters` are going to be renamed as well and possibly changed to `internal`. This is tracked in #300 .

Thank you very much for your contribution. :+1:

Graveen:
Very interesting ! Thx @JVirant

PR #306, opened 2021-06-03
Fix day constant and avoid in-game time glitches

Lotendan:
This fixes the glitches we randomly have in-game when server decides to send a time update packet to the client to resynchronize the clocks. When that happens we used to suddenly see daytime changes.
This also allows the clock with /time to fully count towards 23:59 and not rollback halfway around ~20:00.


NetDwarf:
Wow, what a find. Thank you! :+1:

Graveen:
thank you !

PR #312, opened 2021-06-23
fix Packet encryption for 1.110+ and UDP handling

JVirant:
some code is taken from ~~https://github.com/dol-leodagan/uthgard-opensource~~ https://github.com/thekroko/uthgard-opensource (Metty)
not tested with every versions of game.dll


JVirant:
The UDP part works without the encryption for some recent versions, the client stopped checking if encryption is activated.

NetDwarf:
Very nice find! Thank you! :+1:

PR #313, opened 2021-06-23
fix spell level in delve tooltip/popup

JVirant:
I think we should move the spell level in spell's table instead of having it in the spell line. I don't think a same spell is used in different spell lines with a different level and even if it's the case, the spell can be duplicated.


NetDwarf:
This looks good. Thank you! :+1:

PR #314, opened 2021-06-23
Introduce .Net5 build

NetDwarf:


NetDwarf:
Let's start the move to .Net5. It has still multiple issues, but it does not seem to have obvious issues on the server side (tested locally). 

Known issues: 
- DOLConfig does not work on Linux (we should wait how MAUI turns out, since Linux support seems shaky)
- PerformanceCounter does not work on Linux
(if more comes up I open an issue and collect issues regarding .Net5)

@JVirant raised reasonable concerns regarding locks, however this seems to be fixed by now, but this needs some real world testing to be sure this is not a problem elsewhere (than dotnet/runtime#8255 )

Graveen:
Very nice ! There is also Uno platform that seems better than MAUI for Linux support, but... i don't see this as rock solid as MAUI can be, because i have big concerns about 3rd party techno - related to support.
Kudos !

NetDwarf:
Uno Platform sounds interesting. Let's see how well MAUI for Linux turns out and in the worst case we can use Uno instead maybe. In general I feel like the GUI of DOLConfig is not as important (yet) and it still works for Windows, so I think there is no hurry. :grin:

GameServer/gameutils/Util.cs
```diff
@@ -411,7 +411,7 @@ public static StackTrace GetThreadStack(Thread thread)
 
 			try
 			{
-				trace = new StackTrace(thread, true);
+				trace = new StackTrace(true);
```
JVirant, line 414: This function doesn't work, it should returns the StackTrace from the thread in argument but it's returning the StackTrace of the current thread.

JVirant (commented):
Great job, I will try that when I have some time!  👍

GameServer/gameutils/Util.cs
NetDwarf: `new StackTrace(Thread, bool)` has been deprecated for quite some time without any replacement.

JVirant: If there is no replacement, it's better to throw a NotImplemented (or returning null) than keeping a buggy behavior especially for a debugging tool.

NetDwarf:
Removed integration test for StackTrace, because the output is unstable. 

`Util.GetFormattedStackTraceFrom(Thread)` is on par with `Util.FormatStackTrace(Util.GetThreadStack(Thread))` except missing file and line/column number.

On par also means that it does not work under Linux+Net4.8 (but under .Net5). This was also the case for the old implementation because `new StackTrace(Thread, bool)` throws a PlatformNotSupportedException on Linux.

GameServer/gameutils/Util.cs
NetDwarf: Replaced calls with `Util.GetFormattedStackTraceFrom(Thread)` and using Microsoft.Diagnostics.Runtime for StackTrace output.

NetDwarf:
Okay, if there are no objections I am going to merge it in a day or two.

PR #315, opened 2021-06-23
Fix for Poisoned weapons trade + loot bug as a GM

Lord-Seithenin:
- Players can't re-apply a poison on a weapon beeing already poisonous
- Poisoned Weapons lose their poison if traded with classes other than Infiltrator, Shadowblade, Nightshade
- Fix bug when loot items as GM with "/viewloot random inv"

### DIFF
--- a/GameServer/gameutils/PlayerTradeWindow.cs
+++ b/GameServer/gameutils/PlayerTradeWindow.cs
@@ -619,7 +619,7 @@ namespace DOL.GS
                         InventoryItem itemtoadd = item;
 
                         // If PLayer is not Infiltrator (9), Shadowblade (23), Nightshade (49), remove Envenom bonus before add the item in the inventory
-                        if (item.PoisonSpellID > 0 && !(partner.CharacterClass.ID == 9 || partner.CharacterClass.ID == 23 || partner.CharacterClass.ID == 49))
+                        if (item.PoisonSpellID > 0)
                         {
                             itemtoadd = GameInventoryItem.Create(itemtoadd);
                             itemtoadd.PoisonCharges = 0;
--- a/GameServer/gameutils/PlayerTradeWindow.cs
+++ b/GameServer/gameutils/PlayerTradeWindow.cs
@@ -616,13 +616,24 @@ namespace DOL.GS
 
 						bool tradeSuccess = false;
 
+                        InventoryItem itemtoadd = item;
+
+                        // If PLayer is not Infiltrator (9), Shadowblade (23), Nightshade (49), remove Envenom bonus before add the item in the inventory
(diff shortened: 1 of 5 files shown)


NetDwarf (changes requested):
- `/mob viewloot random inv` works as expected with this patch
- applying poison as proposed is neither livelike nor sensible, since assassins can't change poison at all without changing weapon (relog?)

GameServer/gameutils/PlayerTradeWindow.cs
```diff
@@ -616,13 +616,24 @@ public bool AcceptTrade()
 
 						bool tradeSuccess = false;
 
+                        InventoryItem itemtoadd = item;
+
+                        // If PLayer is not Infiltrator (9), Shadowblade (23), Nightshade (49), remove Envenom bonus before add the item in the inventory
+                        if (item.PoisonSpellID > 0 && !(partner.CharacterClass.ID == 9 || partner.CharacterClass.ID == 23 || partner.CharacterClass.ID == 49))
```
NetDwarf, line 622: Just drop the envenom from weapon regardless of class for consistency. It would be abused either way.

DOL-Avalonia: Fixed

GameServer/gameutils/PlayerTradeWindow.cs
```diff
@@ -648,13 +659,24 @@ public bool AcceptTrade()
 
 						bool tradeSuccess = false;
 
+                        InventoryItem itemtoadd = item;
+
+                        // If PLayer is not Infiltrator (9), Shadowblade (23), Nightshade (49), remove Envenom bonus before add the item in the inventory
+                        if (item.PoisonSpellID > 0 && !(m_owner.CharacterClass.ID == 9 || m_owner.CharacterClass.ID == 23 || m_owner.CharacterClass.ID == 49))
```
Lotendan, line 665: The comment from @NetDwarf applies here too I guess, let's just remove envenom bonus from everyone.

GameServer/gameutils/PlayerTradeWindow.cs
```diff
@@ -616,13 +616,24 @@ public bool AcceptTrade()
 
 						bool tradeSuccess = false;
 
+                        InventoryItem itemtoadd = item;
+
+                        // If PLayer is not Infiltrator (9), Shadowblade (23), Nightshade (49), remove Envenom bonus before add the item in the inventory
+                        if (item.PoisonSpellID > 0)
+                        {
+                            itemtoadd = GameInventoryItem.Create(itemtoadd);
```
Lotendan, line 624: Not sure to understand why we need to recreate a GameInventoryItem from it?

PR #318, opened 2021-06-26
Fix Realm abilities + Add ShadowStrike & Assassinate

Lord-Seithenin:
-Fix Anger of the Gods Ability
-Fix Remedy Ability
-Fix Long Wind Ability
-Add Assassinate RR5 Ability
-Add Shadowstrike RR5 Ability

### DIFF
--- a/GameServer/gameobjects/GamePlayer.cs
+++ b/GameServer/gameobjects/GamePlayer.cs
@@ -57,6 +57,8 @@ namespace DOL.GS
 
 		private readonly object m_LockObject = new object();
 
+        private bool stayStealth = false;
+
 		#region Client/Character/VariousFlags
 
 		/// <summary>
@@ -15953,6 +15955,11 @@ namespace DOL.GS
 					: null;
 			}
 		}
-		#endregion
+
+        /// <summary>
+        /// This property is use for the assassinate RA
+        /// </summary>
+        public bool StayStealth { get => stayStealth; set => stayStealth = value; }
+        #endregion
 	}
 }
--- a/GameServer/gameobjects/GamePlayer.cs
+++ b/GameServer/gameobjects/GamePlayer.cs
@@ -57,8 +57,6 @@ namespace DOL.GS
 
 		private readonly object m_LockObject = new object();
 
-        private bool stayStealth = false;
-
 		#region Client/Character/VariousFlags
 
 		/// <summary>
@@ -15959,7 +15957,7 @@ namespace DOL.GS
         /// <summary>
         /// This property is use for the assassinate RA
         /// </summary>
-        public bool StayStealth { get => stayStealth; set => stayStealth = value; }
+        public bool StayStealth { get; set; }
         #endregion
 	}
 }
--- a/GameServer/realmabilities/handlers/AngerOfTheGodsAbility.cs
+++ b/GameServer/realmabilities/handlers/AngerOfTheGodsAbility.cs
@@ -11,13 +11,49 @@ namespace DOL.GS.RealmAbilities
 	public class AngerOfTheGodsAbility : TimedRealmAbility
 	{
         private DBSpell m_dbspell;
-        private Spell m_spell = null;
+        private Spell m_spell;
         private SpellLine m_spellline;
-        private double m_damage = 0;
+        private double m_damage;
         private GamePlayer m_player;
 
-        public AngerOfTheGodsAbility(DBAbility dba, int level) : base(dba, level) {}
-        public virtual void CreateSpell(double damage)
+        public AngerOfTheGodsAbility(DBAbility dba, int level, DBSpell dbspell) : base(dba, level)
+        {
+            m_dbspell = dbspell;
+        }
+        public AngerOfTheGodsAbility(DBAbility dba, int level) : base(dba, level)
+        {
+            CreateSpell(CalculDomage(level));
+        }
+
+        private double CalculDomage(int level)
+        {
+            double damage = 0;
+            if (ServerProperties.Properties.USE_NEW_ACTIVES_RAS_SCALING)
+            {
+                switch (level)
+                {
+                    case 1: damage = 10.0; break;
+                    case 2: damage = 15.0; break;
+                    case 3: damage = 20.0; break;
+                    case 4: damage = 25.0; break;
+                    case 5: damage = 30.0; break;
+                    default: break;
+                }
+            }
+            else
+            {
+                switch (level)
+                {
+                    case 1: damage = 10.0; break;
+                    case 2: damage = 20.0; break;
+                    case 3: damage = 30.0; break;
+                    default: break;
+                }
+            }
+            return damage;
+        }
+
+        private void CreateSpell(double damage)
         {
             m_dbspell = new DBSpell();
             m_dbspell.Name = "Anger of the Gods";
--- a/GameServer/realmabilities/handlers/RealmAbility.cs
+++ b/GameServer/realmabilities/handlers/RealmAbility.cs
@@ -50,22 +50,15 @@ namespace DOL.GS.RealmAbilities
 		/// <summary>
 		/// max level this RA can reach
 		/// </summary>
-        public virtual int MaxLevel => 0;
-        public virtual int GetValueDelve(int level)
-        {
-            return level;
-        }
-        /// <summary>
-        /// used for 1.110 RA delves
-        /// </summary>        
-        public virtual int AmountPerLevel(int level)
-        {
-            return 0;
-        }
-        /// <summary>
-        /// used for 1.110 RA delves
-        /// </summary>
-        public virtual int GetDurationDelve(int level)
+		public virtual int MaxLevel
+		{
+			get { return 0; }
+		}
+
+		/// <summary>
+		/// used for 1.110 RA delves
+		/// </summary>        
+		public virtual int AmountPerLevel(int level)
         {
             return 0;
         }
--- a/GameServer/realmabilities/handlers/RealmAbility.cs
+++ b/GameServer/realmabilities/handlers/RealmAbility.cs
@@ -50,10 +50,25 @@ namespace DOL.GS.RealmAbilities
 		/// <summary>
 		/// max level this RA can reach
 		/// </summary>
-		public virtual int MaxLevel
-		{
-			get { return 0; }
-		}
+        public virtual int MaxLevel => 0;
+        public virtual int GetValueDelve(int level)
+        {
+            return level;
+        }
+        /// <summary>
+        /// used for 1.110 RA delves
+        /// </summary>        
+        public virtual int AmountPerLevel(int level)
+        {
+            return 0;
+        }
+        /// <summary>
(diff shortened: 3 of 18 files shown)


GameServer/realmabilities/handlers/rr5/AssassinateAbility.cs
```diff
@@ -0,0 +1,75 @@
...
+using DOL.GS.PacketHandler;
+using DOL.GS.Spells;
+using DOL.Language;
+
+namespace DOL.GS.RealmAbilities
+{
+    /// <summary>
+    /// Assassinate inherit of RR5RealmAbility (RealmLevel >= 40 and free)
+    /// </summary>
+    public class AssassinateAbility : RR5RealmAbility
```
NetDwarf, line 11: This is also not an RR5 ability anymore.

GameServer/spells/AssassinateSpellHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+using DOL.Events;
+using DOL.GS.PacketHandler;
+using DOL.GS.Styles;
+using DOL.Language;
+using System;
+
+namespace DOL.GS.Spells
+{
+    [SpellHandler("Assassinate")]
+    public class AssassinateHandler : SpellHandler
```
NetDwarf, line 11: "Could not Instanciate Ability Assassinate from DOL.GS.RealmAbilities.AssasinateAbility".

GameServer/realmabilities/handlers/rr5/ShadowStrikeAbility.cs
```diff
@@ -0,0 +1,80 @@
...
+using DOL.GS.Spells;
+using DOL.Language;
+
+
+namespace DOL.GS.RealmAbilities
+{
+    /// <summary>
+    /// Shadow Strike Ability inherit of RR5RealmAbility (RealmLevel >= 40 and free)
+    /// </summary>
+    public class ShadowStrikeAbility : RR5RealmAbility
```
NetDwarf, line 12: Not an RR5 ability anymore.

NetDwarf (changes requested):
- in general it's better to even split up PRs further since this touches 5 different abilities
- Long Wind fix works
- Shadow Strike is likely not working as intended and is also no longer an ability in the game as far as I know
(Shadowblade – Shadow Strike – Shadowblade takes 10 seconds to disappear in the shadows. Once in the shadows the SB will automatically teleport to and perf any enemy they choose within 1000 radius. Usable only at night. Can not be used to enter a locked keep or tower. 60s duration, 10m re-use.)

GameServer/spells/AssassinateSpellHandler.cs
DOL-Avalonia: As you can see there is a mistake in the table "Abilities" regarding "AssasinateAbility". I've fixed this mistake in the last release of DOL PUBLIC DB here : https://github.com/Eve-of-Darkness/db-public/pull/222/commits/75abb481980ae7884b9b341f7e17528eb16f8737

Lord-Seithenin: As you can see there is a mistake in the table "Abilities" regarding "AssasinateAbility". I've fixed this mistake in the last release of DOL PUBLIC DB here : Eve-of-Darkness/db-public@75abb48

NetDwarf: Thank you for the reply. Missed the typo :sweat_smile:

NetDwarf: I was actually under the impression this is now a regular ability in Stealth skill, but it was actually removed from Live. This makes it harder to reason about this. The ability itself should be casted while being stealthed I think (currently breaks stealth).

GameServer/realmabilities/handlers/rr5/ShadowStrikeAbility.cs
Lord-Seithenin: Fixed, now it's a TimedRealmAbility

GameServer/realmabilities/handlers/rr5/AssassinateAbility.cs
Lord-Seithenin: Fixed, now it's a TimedRealmAbility

GameServer/gameobjects/GamePlayer.cs
```diff
@@ -57,6 +57,8 @@ public class GamePlayer : GameLiving
 
 		private readonly object m_LockObject = new object();
 
+        private bool stayStealth = false;
```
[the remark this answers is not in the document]
Lord-Seithenin: Fixed, line removed

GameServer/realmabilities/handlers/RealmAbility.cs
```diff
@@ -50,10 +50,25 @@ public virtual bool CheckRequirement(GamePlayer player)
...
-		}
+        public virtual int MaxLevel => 0;
+        public virtual int GetValueDelve(int level)
+        {
+            return level;
+        }
+        /// <summary>
+        /// used for 1.110 RA delves
+        /// </summary>        
+        public virtual int AmountPerLevel(int level)
```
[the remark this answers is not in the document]
Lord-Seithenin: This is to display the value per level for the ability in delves

GameServer/realmabilities/handlers/RealmAbility.cs
```diff
@@ -65,13 +80,35 @@ public virtual void AddDelve(ref MiniDelveWriter w)
...
+                    if (AmountPerLevel(i + 1) > 0)
+                        w.AddKeyValuePair(string.Format("AmountLvl_{0}", i + 1), AmountPerLevel(i + 1));
+                }
 			}
 		}
 
+        /// <summary>
+		/// Delve for this RA 1.110+ with gameclient param for multiple delves with one RA. TODO review
+		/// </summary>
+		public virtual void AddDelve(ref MiniDelveWriter w, int level, GameClient clt)
```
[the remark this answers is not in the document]
Lord-Seithenin: This is to display the value per level for the ability in delves

GameServer/realmabilities/handlers/RealmAbility.cs
```diff
@@ -50,10 +50,25 @@ public virtual bool CheckRequirement(GamePlayer player)
...
+        /// used for 1.110 RA delves
+        /// </summary>        
+        public virtual int AmountPerLevel(int level)
+        {
+            return 0;
+        }
+        /// <summary>
+        /// used for 1.110 RA delves
+        /// </summary>
+        public virtual int GetDurationDelve(int level)
```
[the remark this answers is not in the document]
Lord-Seithenin: Fixed and removed, this method doesn't have to be used here

GameServer/realmabilities/handlers/RealmAbility.cs
```diff
@@ -50,10 +50,25 @@ public virtual bool CheckRequirement(GamePlayer player)
 		/// <summary>
 		/// max level this RA can reach
 		/// </summary>
-		public virtual int MaxLevel
-		{
-			get { return 0; }
-		}
+        public virtual int MaxLevel => 0;
+        public virtual int GetValueDelve(int level)
```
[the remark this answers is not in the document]
Lord-Seithenin: Fixed and removed, the method GetValueDelve doesn't have to be used here

GameServer/realmabilities/handlers/AngerOfTheGodsAbility.cs
```diff
@@ -11,13 +11,49 @@ namespace DOL.GS.RealmAbilities
...
+        public AngerOfTheGodsAbility(DBAbility dba, int level, DBSpell dbspell) : base(dba, level)
+        {
+            m_dbspell = dbspell;
+        }
+        public AngerOfTheGodsAbility(DBAbility dba, int level) : base(dba, level)
+        {
+            CreateSpell(CalculDomage(level));
+        }
+
+        private double CalculDomage(int level)
```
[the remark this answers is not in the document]
Lord-Seithenin: We need to create the spell in the constructor with 2 parameters else the server fails to load the ability and to create the spell we need to calculate the damages so it's the reason of the CalculateDamage method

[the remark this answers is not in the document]
Lord-Seithenin: mmh actually I may be wrong... the issue was on my server but seems everything's fine on DOL regarding AngerOfTheGods...

DOL-Avalonia:
> * in general it's better to even split up PRs further since this touches 5 different abilities
> * Long Wind fix works
> * Shadow Strike is likely not working as intended and is also no longer an ability in the game as far as I know
>   (Shadowblade – Shadow Strike – Shadowblade takes 10 seconds to disappear in the shadows. Once in the shadows the SB will automatically teleport to and perf any enemy they choose within 1000 radius. Usable only at night. Can not be used to enter a locked keep or tower. 60s duration, 10m re-use.)

Regarding Shadow Strike Ability, it should still exist on live :
https://darkageofcamelot.com/content/1121e-pendragon-patch-notes

The point is that I need to add some few modifications on my code to add Backstab 2 instead of Perforate Artery for players using ShadowStrike if they have a low specialization in Critical Strike.

GameServer/gameobjects/GamePlayer.cs
```diff
@@ -15953,6 +15953,11 @@ public GamePlayer Bodyguard
 					: null;
 			}
 		}
-		#endregion
+
+        /// <summary>
+        /// This property is use for the assassinate RA
+        /// </summary>
+        public bool StayStealth { get; set; }
```
NetDwarf, line 15960: I'm not sure this is really necessary. Semantically it does not belong here.

DOL-Avalonia: This is the only way I found to make the player using Assassinate skill, keep being invisible after using the skill. If I remove that part, the player will no longer be invisible to other player. Or is it supposed to be what should happen ?

DOL-Avalonia: Regarding Shadowstrike, do you need me to do the patch right now for players to use Backstab 2 instead of Perforate Artery if they have a low specialization in Critical Strike ? Or will you rather accept tu merge this PR and I will do the patch later ?

NetDwarf: For Shadowstrike I would accept it as is. As I see it is now like Assassinate, but with with PA/BS at the end. On Live you need to "cast" the ability while being stealthed. There is no cast animation and the cast time seems to be fixed (10 seconds). 

StayStealth has to be removed. I know it seems minor, however you probably see how big GameLiving/GamePlayer already are and adding another property that belongs to only one ability does not help (in there are more examples for this already :disappointed: ). I'd say even a partly broken Assassinate ability would be better. Maybe you can use the existing `GamePlayer.Stealth(bool)` :cry:  or plain remove Assassinate. You probably need `GamePlayer.Stealth(bool)` for Shadowstrike as well.

PR #317, opened 2021-06-25
add spellhandler's specific TooltipDelve infos for 1.110+

JVirant:
The purpose of this commit is to give a better tooltip for spells with 1.110+.
It's clearly not 100% correct but it's a really big task to do it, I extracted all possible "Function" keywords:
```
absorb
accuracy
add_effect
add_radius
addattack
af_hit
amnesia
araisedead
archery
archmagery
armorpiece_debuff
arrogant
aura
autores
bainshee_pulse
ball
battery
bedazzle
bolt
breath
buff_shear
charm
climb
combat
combat_heal
cost_reduction
cower
critical
crit_buff
damage
damage_type
dashing
dcharm
dconvert
def_proc
degrade
degrade_hurt
direct
direct_chain
direct_inc
disarm
disease
dmg_add
dmg_shield
dmg_suppress
dmg_to_power
doppelgang
dot
dsummon
dummy
efatigue
ehealth
end_drain
enhance_ab
enhancement
epower
fat_heal
fatigue
feather
font
g_absorb
gate
grapple
gsummon
haste
heal
heal_bonus
healsight
hit_buffer
hold_storm
ill_animist
ill_bard
ill_bone
illusion
invisible
item
level
lifedrain
light
lull
mag_pierce
mez_dampen
mod
nabsorb
narchmage
narrogant
nbonus
ndamage
ndisorient
nencumber
nfatigue
nhidden
nillusion
no_cost
nradius
nresist_dam
nresistance
nshield
nskill
nstat
nstat_block
nstat_fade
nstat_perc
nstatdrain
nstyle
ntarget
ntransfer
ntwostat
nullskill
nweaponskill
off_proc
olife_drain
omni_heal
paralyze
petcast
plifedrain
power_heal
power_xfer
powerdrain
preserve
proj_shield
pshield
purge
queue
raise_dead
realmlore
reclaim
recover
regen
regen_perc
rem_eff_ty
rem_rnd_buf
remove_eff
reset_qc
resistance
reveal
rune
rune_sweep
s_predict
sever
shield
siege
siege_protection
skill
snare
snare_dampen
soff_proc
sp_suppress
speedwarp
spell_pulse
spreadheal
ssummon
stat
storm
styleheal
summon
suppress
taunt
teleport
tempest
transfer
true_speed
twostat
vanish
vision
vortex
xrbchp_bonus
xrecover
```

Many of them are not for spells but it can help to find some function used by spells.

PR #319, opened 2021-06-26
Replace Power cost by Health Cost for savage spells in Delves

Lord-Seithenin:
Savage Class use Health instead of power for its spells, as a result, it should be indicated in the Delves.

### DIFF
--- a/GameServer/spells/Savage/SavageBuff.cs
+++ b/GameServer/spells/Savage/SavageBuff.cs
@@ -19,6 +19,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using DOL.GS;
 using DOL.GS.Effects;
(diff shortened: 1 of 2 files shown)


Lotendan:
Hello, thanks for the PR. Is this possible that this is a duplicate of #317 ?

DOL-Avalonia:
Seems there are similarities yes, but this commit has been developed in a different project without knowing #317 would be proposed. My focus was to fix the Delves infos of the savage only.

Lord-Seithenin:
Seems there are similarities yes, but this commit has been developed in a different project without knowing #317  would be proposed. My focus was to fix the Delves infos of the savage only.

Lord-Seithenin:
I will actually remove this PR, because the results are similar in PR #317

PR #320, opened 2021-06-26
Fix Pets following player erratically

Lord-Seithenin:
- Now Pets will follow players in relation with their speed
- In order to make this work, some adaptations were necessary on BomberBrains while in combat in order to hit their target normally (without beeing stuck)

### DIFF
--- a/GameServer/gameobjects/GameNPC.cs
+++ b/GameServer/gameobjects/GameNPC.cs
@@ -1747,7 +1747,7 @@ namespace DOL.GS
             if (InCombat || Brain is BomberBrain)
                 WalkTo(newX, newY, (ushort)newZ, MaxSpeed);
             else if (MaxSpeed < GetDistance(new Point2D(newX, newY)))
-                WalkTo(newX, newY, (ushort)newZ, Math.Min(MaxSpeed, followLiving.CurrentSpeed + 50));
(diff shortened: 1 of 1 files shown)


Lotendan:
Hi there, nice fix.
Tested in my local, and this works like a charm!
One remark though: shouldn't we give the pet a slightly higher speed than the player just to allow it to catch up in case it gets too far? Something like `followLiving.CurrentSpeed + 50`.
Also you seem to handle the case of `BomberBrain` specifically. What about theurg pets though? Can we consider they will always be in combat?

Lotendan (approved):

Lord-Seithenin:
You're right, the pet may appear a bit too slow so I may improve it a bit like you said.
Regarding the bomber they are supposed to go directly toward the player then explode on it (and be destroyed in the process), the issue was that those bombers were accumulating in a specific spot when/if the player was moving while targeted by the caster, and then if the targeted player suddently stopped to move all the bombers which accumulated would suddently explode alltogether on the same time on their target. You can try to remove the "bomber part" from your project and see what would happen...

Lotendan (approved):

NetDwarf (approved):
Works as intended.

Thank you! :+1:

PR #321, opened 2021-06-26
fix udp: refactor CombinePackets and change the UDP send buffer to 512

JVirant:


JVirant:
Actually, I don't know if it's a bug from the client, the logger or the server but if an udp packet is combined with another one, the client starts to ignore some updates and the logger give packets with code 00 :
```
<RECV UDP Time: 00h00m28s922ms Code:0x0000 Len: 33>
00 37 26 24 00 00 C4 99 00 00 0C 10 00 00 06 94   .7&$..ę.......�
00 00 64 00 00 18 00 22 A1 00 00 00 31 27 24 00   ..d...."�...1'$.
00                                                .
```

From my understanding from the reverse of the client it should be ok to combine some packets... Since an udp packet has way more chance to be received if it's small, I think it's better to not combine them.

GameServer/packets/Server/GSUDPPacketOut.cs
```diff
@@ -54,12 +54,6 @@ public override ushort WritePacketLength()
 		{
 			Position = 0;
 			WriteShort((ushort)(Length-5));
-
-			//IMPORTANT!!!
-			//Set the capacity of the internal buffer or
-			//the byte-array of GetBuffer will be TOO big!
-			Capacity = (int)Length;
```
Lotendan, line 61: From the comments this looks important. Shouldn't we keep this? Or what is the effect of removing it?

Lotendan (commented):
I guess this simplifies a lot, thanks for the change.

GameServer/packets/Server/GSUDPPacketOut.cs
JVirant: No, it's not needed anymore since I use the length from the packet's data so it save a buffer reallocation.

Lotendan (approved):

Lotendan (approved):
Thanks a lot for the PR.
Personally I think I would AcquirePacketBuffer for UDP as well and not create a new smaller buffer, but that works too.

JVirant:
> 
> 
> Thanks a lot for the PR.
> Personally I think I would AcquirePacketBuffer for UDP as well and not create a new smaller buffer, but that works too.

The AcquirePacketBuffer is more useful to create a packet buffer than a sending buffer which can be way bigger and in all case, it just do a "new byte[]".

NetDwarf:
I can't claim to understand it fully (roughly at best) and therefore possibly not appreciate it for the right things, however beside not obviously breaking anything, 1.127d client behaves way better (fewer packet losses, no npc blinking on test server).

If there are no contraindications I merge it tomorrow. :smiley: 

Thank you! :+1:

PR #322, opened 2021-06-29
add basic 1.127 support, RC4 property and some fix 1.124+

JVirant:
I fixed some issue that I created from previous commits especially with the encryption.

### DIFF
--- a/GameServer/packets/Server/PacketLib1124.cs
+++ b/GameServer/packets/Server/PacketLib1124.cs
@@ -16,7 +16,6 @@
 * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 *
 */
-#define NOENCRYPTION
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -1568,7 +1567,7 @@ namespace DOL.GS.PacketHandler
 						else if ((texture & 0xFF) != 0)
 							pak.WriteByte((byte)texture);
 						if (item.Effect != 0)
-							pak.WriteShort((byte)item.Effect); // effect changed to short
+							pak.WriteShort((ushort)item.Effect); // effect changed to short
 					}
 				}
 				else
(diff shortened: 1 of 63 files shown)


GameServer/serverproperty/ServerProperties.cs
```diff
@@ -121,6 +121,12 @@ public static void InitProperties()
 		[ServerProperty("system", "client_version_max", "Maximum Client Version - Edit this to change which client version at the most have to be used: -1 = any, 1.80 = 180", -1)]
 		public static int CLIENT_VERSION_MAX;
 
+		/// <summary>
+		/// Enable RC4 encryption for communication between the client and the server
+		/// </summary>
+		[ServerProperty("system", "client_enable_encryption_rc4", "Enable client RC4 encryption - Advanced user only, you need a special launcher to enable encryption", false)]
```
Lotendan, line 127: I was actually worried about the launcher, but you mention it properly here ;)

GameServer/packets/Server/PacketLib1124.cs
```diff
@@ -3134,80 +3133,34 @@ public virtual void SendRealm(eRealm realm)
 				SendTCP(pak);
 			}
 		}
-		public virtual void SendRegions()
+		public virtual void SendRegions(ushort regionId)
```
Lotendan, line 3136: Seeing that we are only sending one region now, should we make this singular?
`SendRegion`?

JVirant: I stay conservative about changes on public methods to keep a better compatibility with existent scripts or custom code.

NetDwarf:
Works like a charm. Thank you! :smiley:

JVirant:
> Works like a charm. Thank you! 😃

It doesn't... :(
I got a crash when I was with 2 clients : one 1.125d and another one 1.127d but it's a start!

NetDwarf:
I should maybe say works like advertised :smile:

PR #324, opened 2021-07-01
Automatically equip left-handed items

Lotendan:
Here we consider items that are Item_Type = LeftHandWeapon and Hand = 2.
When moved from any slot onto the player paper doll, they were given a toSlot = LeftHand.
This would work for characters with ability to wield left handed weapons.
But if the character does not have this ability, the object would simply tell "This can't go there.".

The proper solution is to always put "Usable in left hand" weapons in the right hand weapon.
I verified on live and this is the expected behaviour.
If you want to put a weapon in the left slot, you have to tell it explicitely by putting it on the left slot and not on the paper doll.


NetDwarf:
Works as expected. Thank you :+1:

PR #327, opened 2021-07-07
Defensive opener fix

kbvandergriffjr:


GameServer/packets/Client/168/DetailDisplayHandler.cs
```diff
@@ -2129,8 +2131,16 @@ public static string DelveStyle(GameClient clt, int id)
 				//.Value("Skill",43)
 				if (style.GrowthRate>0)
 					dw.AddKeyValuePair("OpeningDamage",style.GrowthRate*100);
-				//.Value("SpecialValue", GetSpecialValue(style),GetSpecialValue(style)!=0)
-				//.Value("FollowupStyle",style.DelveFollowUpStyles,!string.IsNullOrEmpty(style.DelveFollowUpStyles))
+                //.Value("SpecialValue", GetSpecialValue(style),GetSpecialValue(style)!=0)
+                //.Value("FollowupStyle",style.DelveFollowUpStyles,!string.IsNullOrEmpty(style.DelveFollowUpStyles))
+                if (style.Procs != null && style.Procs.Count > 0)
+                {
+                    foreach (Tuple<Spell, int, int> proc in style.Procs)
```
Lotendan, line 2138: Can we really have multiple tooltips in the proc?
Let's imagine we have multiple `style.Procs` where `style.Procs.Count > 1`.
In this case you will loop over the different procs and in the end the `MiniDelveWriter` which ensures the unicity of the key-values will erase all the previous procs and you will end up with only the last proc in the list being written in the tooltip.
Here, suggest to take the first or the last proc in the list: `style.Procs.FirstOrDefault()`.

GameServer/packets/Client/168/DetailDisplayHandler.cs
```diff
@@ -2129,8 +2131,16 @@ public static string DelveStyle(GameClient clt, int id)
...
-				//.Value("SpecialValue", GetSpecialValue(style),GetSpecialValue(style)!=0)
-				//.Value("FollowupStyle",style.DelveFollowUpStyles,!string.IsNullOrEmpty(style.DelveFollowUpStyles))
+                //.Value("SpecialValue", GetSpecialValue(style),GetSpecialValue(style)!=0)
+                //.Value("FollowupStyle",style.DelveFollowUpStyles,!string.IsNullOrEmpty(style.DelveFollowUpStyles))
+                if (style.Procs != null && style.Procs.Count > 0)
+                {
+                    foreach (Tuple<Spell, int, int> proc in style.Procs)
+                    {
+                        dw.AddKeyValuePair("SpecialNumber", proc.Item1.InternalID);
+                        dw.AddKeyValuePair("SpecialType", 1);
```
Lotendan, line 2141: `SpecialType` value depends on the spell which is actually put in the proc.
For this reason I believe having a `TooltipDelveStyle` override in `SpellHandler` for various spells would make sense as each of the spells has its own way to delve itself when put in a style.

GameServer/skillhandler/TripleWieldAbilityHandler.cs
```diff
@@ -23,7 +23,7 @@ public class TripleWieldAbilityHandler : IAbilityActionHandler
 		/// <summary>
 		/// The ability reuse time in seconds
 		/// </summary>
-		protected const int REUSE_TIMER = 7 * 60; // 7 minutes 
+		protected const int REUSE_TIMER = 10 * 60; // clait: 10 minutes [og: 7]
```
Lotendan, line 26: I don't know about the others but I just checked on live now and Triple Wield is still 7 minutes, so I wouldn't go about changing it here.

GameServer/skillhandler/StagAbilityHandler.cs
```diff
@@ -41,7 +41,7 @@ public class StagAbilityHandler : IAbilityActionHandler
 		/// <summary>
 		/// The ability reuse time in milliseconds
 		/// </summary>
-		protected const int REUSE_TIMER = 60000 * 20;
+		protected const int REUSE_TIMER = 60000 * 10; // clait: 10 minutes [og: 20]
```
Lotendan, line 44: You do not need to put your name in the comments ;) `git blame` is here for that exact purpose ;)

GameServer/spells/Animist/SummonAnimistFnF.cs
```diff
@@ -80,10 +80,11 @@ public override void ApplyEffectOnTarget(GameLiving target, double effectiveness
 			(m_pet.Brain as TurretBrain).IsMainPet = false;
 
 			(m_pet.Brain as IOldAggressiveBrain).AddToAggroList(target, 1);
-			(m_pet.Brain as TurretBrain).Think();
+			// (m_pet.Brain as TurretBrain).Think(); clait: Moving after the PetCount increase to fix Turrent count TODO Review
 			//[Ganrod] Nidel: Set only one spell.
 			(m_pet as TurretPet).TurretSpell = m_pet.Spells[0] as Spell;
 			Caster.PetCount++;
+			(m_pet.Brain as TurretBrain).Think(); // clait: remove this line to revert
```
Lotendan, line 87: I believe this can even be removed.
A new brain whenever created will immediately `Think()`.

PR #334, opened 2021-09-16
Fix Healer NPC : the healer now cures the RVR ressurection illness

DigitalBox98:
The healer now cures the RVR ressurection illness in addition to the PVE ressurection illness

### DIFF
--- a/GameServer/gameobjects/CustomNPC/Healer.cs
+++ b/GameServer/gameobjects/CustomNPC/Healer.cs
@@ -34,7 +34,8 @@ namespace DOL.GS
 	[NPCGuildScript("Healer")]
 	public class GameHealer : GameNPC
 	{
-		private const string CURED_SPELL_TYPE = "PveResurrectionIllness";
+		private const string PVE_CURED_SPELL_TYPE = "PveResurrectionIllness";
+        private const string RVR_CURED_SPELL_TYPE = "RvrResurrectionIllness";
 
         private const string COST_BY_PTS = "cost";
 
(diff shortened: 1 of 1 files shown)


NetDwarf:

I hope I am not misunderstanding this, but on Live healer (NPC) does not heal RvR sickness. RvR sickness is supposed to be a punishment I think (in the olden days it was the portal ceremony or the long run to the mile gates). If one could just heal it from an NPC it would take away that punishment.

DigitalBox98:
Tested on live and indeed the healer doesn't cure RvR sickness.
So maybe a possibility is to tune the timer relative to the battleground :)  (1 min for the battleground level 10 for instance on live)

DigitalBox98:
As it's not a bug, but a feature I'm closing it :)

NetDwarf:
Nevertheless thank you for your contribution, since you took the time to read, understand and write the change. :+1: 

I also find the suggestion interesting for sickness to be shorter for lower levels or depending on RvR-zone (BG vs open frontier).

JVirant:
It can be a good idea to keep the possibility to have the healer curing the RVR ressurection illness behind a server property.
I think many servers want to change that! 🙂

DigitalBox98:
While closing this PR, I was thinking about that property possibility too :)
I think it shouldn’t be quite complicated to add this property, and I will propose another PR if this feature is interesting (for the moment, I have kept this behavior on my server but without a property)

NetDwarf:
It is better to keep this behavior custom. The shorter RvR sickness in BGs seems to be livelike (1 minute instead of 3) and would be a better solution in general as it does not take away the penalty (only makes it shorter in BGs).

PR #335, opened 2021-09-17
Fix hover descriptions for 1.110+ clients

NetDwarf:
Fix Style damage classification display
Add Style evade/parry/block requirement display
Add Style proc display
Fix Augmented amount of stat RAs delve

Refactorings/Cleanup:
Cleanup noise comments
Remove MiniDelveWriter
Add SpellHandler.ShortDescription property
Move Delves to own class
Remove redundant code from PacketLib1124


NetDwarf:
This is still not exhaustive and has a lot of room for refinement.

Known issues (among others): 
* Ability delves are hard coded, so they need the index the client wants
* RealmAbilties are not correctly delved in general except Augmented stats for the most part (active RAs use legacy delve on right click and have 3 different delve displays for some reason)
* Tireless is not an ability but a realm ability (=> does not delve)

This also needs refinement because this implements parallel hierarchy which is a code smell. For now this is a compromise to get code out of god classes (SpellHandler for example). Altho this is meant to separate abstraction levels, it's still not satisfactory and I am not yet sure how to fix that. 

A next step would be moving the legacy delves out of relevant classes. For me it would make the most sense to change legacy delves to a similar or same display to 1.110+ delves for parity.

PR #336, opened 2021-09-22
ServerListUpdate : use standard UrlEncode from library

DigitalBox98:
ServerListUpdate : use standard UrlEncode from library to handle all special characters

Previous method was only handling the space character.
See https://www.w3schools.com/tags/ref_urlencode.ASP for all the possible cases.


Lotendan (approved):

NetDwarf:
Thank you!

PR #337, opened 2021-09-22
fix text encoding

JVirant:
Daoc use Windows-1252 as text encoding which is the default encoding in .net 4.8 on windows but it's utf-8 for other platforms (and .net core).

### DIFF
--- a/DOLBase/Constants.cs
+++ b/DOLBase/Constants.cs
@@ -100,6 +100,6 @@ namespace DOL
 		/// <summary>
 		/// The default encoding to use for all string operations in packet writing or reading.
 		/// </summary>
-		public static readonly Encoding DefaultEncoding = Encoding.GetEncoding(1252);
+		public static readonly Encoding DefaultEncoding = CodePagesEncodingProvider.Instance.GetEncoding(1252);
 	}
 }
--- a/DOLBase/Constants.cs
+++ b/DOLBase/Constants.cs
@@ -100,6 +100,6 @@ namespace DOL
 		/// <summary>
 		/// The default encoding to use for all string operations in packet writing or reading.
 		/// </summary>
-		public static readonly Encoding DefaultEncoding = Encoding.Default;
+		public static readonly Encoding DefaultEncoding = Encoding.GetEncoding(1252);
 	}
 }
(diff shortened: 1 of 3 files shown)


JVirant:
Maybe we can default to ISO-8859-1 which is already included in dotnet...

NetDwarf:
Could you tell me a scenario where this currently causes issues? (also to replicate)

JVirant:
Just speak in french or name a npc with an accent and check the name in game (with dotnet core/5.0)

Lotendan:
I confirm accents causes issues and changing with windows-1252 encoding solves all issues.

NetDwarf:
Is it possible to use ISO-8859-1 (codepage=28591) instead? I am just asking because this would avoid another package. :slightly_smiling_face:

JVirant:
> 
> 
> Is it possible to use ISO-8859-1 (codepage=28591) instead? I am just asking because this would avoid another package. 🙂
> 
> (edit) Rephrased: Is there anything that would prevent us from using ISO-8859-1? I vote for it then.

It's possible to use ISO-8859-1, I think there is only one issue with it.

It's in case of an unknown character of ISO-8859-1 like € received from the client, it will probably throw instead of accepting it. Even if the client doesn't display it correctly (we have a box instead like □).

Lotendan:
Honestly I would prefer if we used windows-1252 straight away so we don’t have to worry about encoding any further or risk getting undetectable bugs buried deep within the client just because of a wrong character.
Plus what is the cost of a package nowadays? This is the most official package you can find, fully bundled and maintained by Microsoft so the risk with maintenance is close to zero.

NetDwarf:
> 
> It's possible to use ISO-8859-1, I think there is only one issue with it.
> 
> It's in case of an unknown character of ISO-8859-1 like € received from the client, it will probably throw instead of accepting it. Even if the client doesn't display it correctly (we have a box instead like □).

Just quick checked with "€" (I can't enter any fancy signs in Windows beside that):
Live reply: `80`
DOL(1252) reply: `80`
DOL(ISO-8859-1) reply: `80`

Given that the client doesn't really understand the Euro sign in the first place makes me doubt Live uses Windows-1252 anyway.

DigitalBox98:
Working fine. Thanks for this fix 👍

PR #338, opened 2021-09-22
remove useless SQL prepare statement

JVirant:
The prepare statement is useful only if we re-use the prepared statement but in DOL, a new statement is created every time.


Lotendan (approved):
Not tested it but this makes sense

NetDwarf:
Yeah this should work (maybe even slightly better). Thank you!

PR #339, opened 2021-09-23
Fix character creation and realm switching in 1.126+

NetDwarf:
Fix character creation, which was previously not usable with other server rules than PvE/PvP
Fix realm switcher not being available after logging a char in and out
Fix a small calculation error in PacketLib1126 weapon damage calc

### DIFF
--- a/GameServer/packets/Server/PacketLib1126.cs
+++ b/GameServer/packets/Server/PacketLib1126.cs
@@ -31,11 +31,21 @@ namespace DOL.GS.PacketHandler
 	{
 		private static readonly ILog log = LogManager.GetLogger(MethodBase.GetCurrentMethod().DeclaringType);
 
-		/// <summary>
-		/// Constructs a new PacketLib for Client Version 1.126
-		/// </summary>
-		/// <param name="client">the gameclient this lib is associated with</param>
-		public PacketLib1126(GameClient client)
+        protected override byte ServerTypeID
+        {
+			get
+			{
+				var serverType = GameServer.Instance.Configuration.ServerType;
+				switch (serverType)
+				{
+					case eGameServerType.GST_PvP: return 1;
+					case eGameServerType.GST_PvE: return 3;
+					default: return 7; //Ywain; 0x00 no longer working
+				}
+			}
+		}
+
+        public PacketLib1126(GameClient client)
 			: base(client)
 		{
 		}
--- a/GameServer/packets/Server/PacketLib1126.cs
+++ b/GameServer/packets/Server/PacketLib1126.cs
@@ -102,9 +102,10 @@ namespace DOL.GS.PacketHandler
 
 			int firstSlot = (byte)realm * 100;
 
+			var enableRealmSwitcherBit = GameServer.ServerRules.IsAllowedCharsInAllRealms(m_gameClient) ? 1 : 0;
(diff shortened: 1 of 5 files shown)


NetDwarf:
Ugh, this one is out of order.

GameServer/packets/Server/PacketLib1126.cs
```diff
@@ -340,10 +348,10 @@ public override void SendUpdateWeaponAndArmorStats()
 				pak.WriteByte(0x00); //unk
 
 				// weapondamage
-				var wd = (int)(m_gameClient.Player.WeaponDamage(m_gameClient.Player.AttackWeapon) * 100.0);
-				pak.WriteByte((byte)(wd / 100));
+				var wd = (int)(m_gameClient.Player.WeaponDamage(m_gameClient.Player.AttackWeapon) * 100);
+				pak.WriteByte((byte)(wd / 0x100));
```
Lotendan, line 352: `0x100` ? Hexa ?

GameServer/packets/Server/PacketLib1126.cs
```diff
@@ -92,9 +102,10 @@ public override void SendCharacterOverview(eRealm realm)
 
 			int firstSlot = (byte)realm * 100;
```
Lotendan, line 104: Cast to `uint` immediately with `u`suffix after the number.

Lotendan (approved):

GameServer/packets/Server/PacketLib1126.cs
NetDwarf: Yes, it is to split the "short" value into separate bytes.

GameServer/packets/Server/PacketLib1126.cs
NetDwarf: I understand the decision here, since the value can at no point overflow. That being said however, this is unrelated to this PR.

Lotendan: Ok wrong line sorry. The comment was meant to apply to line 105 below with the realm bit.

GameServer/packets/Server/PacketLib1126.cs
Lotendan: In this case use bit operations instead of a costly division (shift + AND)

GameServer/packets/Server/PacketLib1126.cs
NetDwarf: It is not enough to warrant a new commit, but it is indeed nicer.

GameServer/packets/Server/PacketLib1126.cs
NetDwarf: Division takes one cycle on almost any CPU (latency is higher tho), but even if it would take 50 cycles it would make no perceivable difference, because it is called rather seldom. For readability I chose division/modulo over `>>`/`&`.

JVirant: I think there is no point of optimization here, the compiler should have no issue to make that as fast as a bitwise operation.

Just keep the most readable version since it's about bits (keeping the high part of the int16), i find bitwise operations more relevant but it's just my opinion.

Graveen: Agree this kind of optimization is normally applied by the compiler whatever form it takes.

PR #340, opened 2021-09-24
Fix DB access to MariaSQL 10.6+

NetDwarf:
Set result character set to UTF8mb4


Lotendan:
From the internet it seems you can’t fix that in the MySQL client directly but need a fix server side.

NetDwarf:
That change does not alter the database and only changes the resulting character set. UTF8mb4 is a strict superset of UTF8mb3 (it otherwise defaults to and triggers the issue), which means it should not change the result. The "explicit" MySQL tests run fine with a 10.6 database and DOL server is also not producing any artifacts in my admittedly limited testing.

NetDwarf:
Apparently NUnit marks all tests that are dependent on a SetupFixture as failed disregard if it is valid according to the given filter or not.

PR #344, opened 2021-10-04
Tooltip fix : Fix the spell level tooltip information

DigitalBox98:
Fix the spell level tooltip information on 1.125 client

This PR is subject to discussion as I've taken code from LosOjos repo where the tooltip level is displayed correctly on 1.125 client.

### DIFF
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -3972,7 +3972,148 @@ namespace DOL.GS.Spells
 			return 0;
 		}
 		#endregion
-		
+
+		/// <summary>
+		/// Return the given Delve Writer with added keyvalue pairs.
+		/// </summary>
+		/// <param name="dw"></param>
+		/// <param name="id"></param>
+		public virtual void TooltipDelve(ref MiniDelveWriter dw, int id, GameClient client)
+		{
+			if (dw == null)
+			{
+				return;
+			}
+			int level = Spell.Level;
+			int spellID = Spell.ID;
+
+			foreach (SpellLine line in client.Player.GetSpellLines())
+			{
+				Spell s = SkillBase.GetSpellList(line.KeyName).Where(o => o.ID == spellID).FirstOrDefault();
+				if (s != null)
+				{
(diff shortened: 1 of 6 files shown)


NetDwarf:
What else beside spell level is not showing in Tooltips?

DigitalBox98:
It seems the Target also is always set to "Self"

DOL:
DOL

Official:
official
.

GameServer/spells/SpellHandler.cs
```diff
@@ -3970,9 +3970,150 @@ public virtual void OnEffectRestored(GameSpellEffect effect, int[] vars)
...
+		}
 		#endregion
-		
+
+		/// <summary>
+		/// Return the given Delve Writer with added keyvalue pairs.
+		/// </summary>
+		/// <param name="dw"></param>
+		/// <param name="id"></param>
+		public virtual void TooltipDelve(ref MiniDelveWriter dw, int id, GameClient client)
```
NetDwarf, line 3981: This makes SpellHandler responsible for packet handling/logic.

NetDwarf (changes requested):
This PR reintroduces a lot of stuff that was removed in #335.

NetDwarf:
> It seems the Target also is always set to "Self"
> 
> DOL: DOL
> 
> Official: official .

Did you check that the entry in DB is indeed correct (Target being realm)?

DigitalBox98:
I have double checked and no issue on the "Target" field from the DOL build :) But unfortunately it seems to be relative to this PR :/

So the only point I have noticed is the level always set to 1. 
Not sure to understand where this kind of fix can be done, but I will read again the various comments :)

NetDwarf:
I can give a two small pointers. Read the description of `SkillBase.GetSpellByTooltipID(objectId)` and look at how `DetailDisplayHandler.DelveSpell` is different from `SpellDelve` constructor.

DigitalBox98:
Ok I have identified the piece of code and proposed another simplier solution with #347 
I don't know if it's the best solution but it works :)

PR #346, opened 2021-10-04
Fix Style proc display

NetDwarf:
Send proc spell description
Change order so procs are sent before the style


NetDwarf:
This bug was a regression that was reintroduced due to the removal of preloading the descriptions. The same applies to the level not showing correctly, however this one needs a fix at a different place.

Issue #263, opened 2020-11-15
1.125 Client spell tool tips


Scrubtasticx:
On client 1125 all the spell tooltips say all spells are level 1 and they don't say how much damage or runspeed % they do now.

Combat skills are fine its just spells

P.S. They work on Los-Ojo's repo say maybe somthing was missed on merge?

tegstewart:
I've noticed some weirdness with multiple spells at the same level not working on 1.125 as well.

DigitalBox98:
Same behavior noticed on spells tooltip :)

PR #347, opened 2021-10-04
Spell tooltip fix : Retrieve the correct spell level

DigitalBox98:
Retrieve the correct spell level to show in the tool tip.

This fix is simplier than the previous PR

### DIFF
--- a/GameServer/gameutils/Skill/Delve/SongDelve.cs
+++ b/GameServer/gameutils/Skill/Delve/SongDelve.cs
@@ -19,6 +19,7 @@
 using DOL.GS.PacketHandler;
 using DOL.GS.Spells;
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


NetDwarf:
This is actually well done! It does exactly what it says on the tin.

Do you want to fix Songs (Level) as well while you're at it? :slightly_smiling_face:

DigitalBox98:
Ok got it working for Song too ;)

There might be some cleanup required in the db-public database because of some duplicate SpellId
I've got a wrong level 1 for a song on the minstrel but it was due to duplicate data in the DB, so so far so good.

spellid

NetDwarf:
Works like a charm! Thank you!

This closes #263. (If there are still issues with hover descriptions a new issue can be opened anyway)

PR #352, opened 2021-10-19
Hotfix: DOLConfig silently failing

NetDwarf:


NetDwarf:
It ain't pretty, but it should do what I want it to do :sweat_smile:

PR #356, opened 2021-10-23
Replace MySql.Data with MySqlConnector

NetDwarf:
Fix MariaDB 10.6+ not working
Fix utf8 databases with special characters


Lotendan:
So it is working on my local. I’ve tried with Maria 10.6 in both read and write mode for special characters and it worked properly. I also tried various other charsets for the tables and they are working as well although I had one small issue with some Chinese character in mb3 but that’s expected I guess. For the rest of the characters they render properly in game to the limits of the chat’s encoding obviously.

I did not manage to re-install a Maria 10.4 without breaking all my config so I was not able to test it.

PR #360, opened 2021-11-06
Fix SQlite bundle for e_sqlite lib using Mono

dol-leodagan:


dol-leodagan:
This is not the reason for having problem with SQLite using Linux and Mono.

SQlite Native binaries are not distributed with Windows Build Release of DOL.

PR #361, opened 2021-11-06
Add SQLite Binaries for Linux / OSX to Windows Release

dol-leodagan:
Improves Release package Compatibility with Mono.

This will allows to release .NetFramework Linux based Docker Images using Release Build artifacts.


NetDwarf:
Thank you! I am going to merge this, but we might want to split binaries again to keep the size down. Additionally we now have the mechanism to port to a Linux base image.

PR #365, opened 2021-11-17
Move to .Net6

NetDwarf:
Rename occurrences of Net5 to NetCore
Update packages:
  System.Text.Encodings.CodePages 5.0.0 -> 6.0.0
  System.Resources.Extensions 5.0.0 -> 6.0.0
  System.ServiceProcess.ServiceController 5.0.0 -> 6.0.0
  System.Diagnostics.PerformanceCounter 5.0.1 -> 6.0.0
  Microsoft.CodeAnalysis.CSharp 3.11.0 -> 4.0.1
  Microsoft.NET.Test.Sdk 16.11.0 -> 17.0.0
  Microsoft.Data.Sqlite 5.0.11 -> 5.0.12
  log4net 2.0.12 -> 2.0.13
  MySqlConnector 1.3.14 -> 2.0.0
  NUnit3TestAdapter 4.0.0 -> 4.1.0


NetDwarf:
Unfortunately I was not able to update Microsoft.Data.Sqlite to 6.x because it would break the Framework builds (can't load libe_sqlite3.dll). There is however no obligation to upgrade as far as I can tell, so there is no harm in using the older version for now.

PR #370, opened 2021-11-24
Fix yell angle

Lotendan:
Yell heading does not depend on target heading but only on current player's heading.
Also used this opportunity to improve the code using `IsSpammingCommand`

### DIFF
--- a/GameServer/commands/playercommands/yell.cs
+++ b/GameServer/commands/playercommands/yell.cs
@@ -16,6 +16,7 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  *
  */
+
 using DOL.GS.PacketHandler;
 
 namespace DOL.GS.Commands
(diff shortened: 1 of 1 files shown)


GameServer/commands/playercommands/yell.cs
```diff
@@ -30,19 +31,12 @@ public class YellCommandHandler : AbstractCommandHandler, ICommandHandler
...
-			const string YELL_TICK = "YELL_Tick";
-			long YELLTick = client.Player.TempProperties.getProperty<long>(YELL_TICK);
-			if (YELLTick > 0 && YELLTick - client.Player.CurrentRegion.Time <= 0)
-			{
-				client.Player.TempProperties.removeProperty(YELL_TICK);
-			}
-
-			long changeTime = client.Player.CurrentRegion.Time - YELLTick;
-			if (changeTime < 750 && YELLTick > 0)
+			if( IsSpammingCommand( client.Player, "yell", 750 ) )
```
NetDwarf, line 34: That is way more elegant than TempProperty. :+1:

NetDwarf:
Nice find, thank you!

PR #372, opened 2021-12-02
Improve SpellHandler to perform repeated checks "during" cast for long running spells

Lotendan:
At the moment the spell handler only performs 3 checks. This is okay for most casts which are shorter than 5 seconds total.
For other spells which are longer, this will be a problem since a lot of things can happen in the meantime of each check.

Let’s take the example of a cast which lasts 15 seconds: `time = 15000`.
`step1 = time / 3` i.e. step 1 = 5000 but with the current values for the server properties step1 is limited to 1500 so step1 = 1500.
Same thing for step3 = 1500.
It follows `step2 = time - step1 - step3` so step2 = 12000.

What this means is the following: CheckAfterCast is executed after 1500ms
CheckDuringCast is executed 12000ms later
CheckEndCast executed 1500ms after.

A lot of things can happen between AfterCast and DuringCast in a 12000ms interval. It would be nice to be able to check more often (i.e. enemy goes out of view, etc…)

Hence the proposal to limit the duration of step2 and to repeat CheckDuringCast several times while the cast is running. This code only impacts long running casts.

“Several times” depend on the value of the new server properties added in this PR. This code will try to subdivide the 12000 ms into the closest integer duration which matches the server properties.

### DIFF
--- a/GameServer/serverproperty/ServerProperties.cs
+++ b/GameServer/serverproperty/ServerProperties.cs
@@ -2287,9 +2287,12 @@ namespace DOL.GS.ServerProperties
 		[ServerProperty("spells", "spell_interrupt_again", "", 100)]
 		public static int SPELL_INTERRUPT_AGAIN;
 
-		[ServerProperty("spells", "spell_interrupt_maxstagelength", "Max length of stage 1 and 3, 1000 = 1 second", 1500)]
+		[ServerProperty( "spells", "spell_interrupt_maxstagelength", "Max length of stages 1 and 3, 1000 = 1 second", 1500 )]
 		public static int SPELL_INTERRUPT_MAXSTAGELENGTH;
-		
+
+		[ServerProperty( "spells", "spell_interrupt_max_intermediate_stagelength", "Max length of stage 2, 1000 = 1 second. 999999 to disable", 3000 )]
+		public static int SPELL_INTERRUPT_MAX_INTERMEDIATE_STAGELENGTH;
+
 		[ServerProperty("spells", "spell_charm_named_check", "Prevents charm spell to work on Named Mobs, 0 = disable, 1 = enable", 1)]
 		public static int SPELL_CHARM_NAMED_CHECK;
 
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -29,6 +29,7 @@ using DOL.GS.PacketHandler;
 using DOL.GS.ServerProperties;
 using DOL.GS.RealmAbilities;
 using DOL.GS.SkillHandler;
+using DOL.GS.Utils;
 using DOL.Language;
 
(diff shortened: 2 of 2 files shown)


NetDwarf:
What problem, that previously existed, is it solving?

Lotendan:
Updated the description above. Hopefully this is clearer :)

NetDwarf:
I made some quick tests with a 15 second casttime and can't see direct issues at any given time in the cast nor cast animation. Could you give a more concrete example of an event/check that is not working in between these intervals?

Given that these are known at spell start (casttime should not be changed while casting) it would also make sense to make these intervals adjusted for the given casttime (i.e. 15 seconds effective cast time, every interval is 5000ms long).

But first things first, we need to find a direct issue with current implementation (beside the obvious quirkiness of the implementation, but that requires a bigger refactoring I think).

Lotendan:
Let’s imagine a DD with target “Enemy” and a 15 seconds cast time.
Target needs to remain within range of the spell and within sight of the caster until the end of the cast.

Caster starts to cast a spell on target at time T. Now target moves out of range during step2 (between CheckAfterCast and CheckDuringCast i.e between T+1.5sec and T+13.5sec). Caster will continue casting the spell for a long time until CheckDuringCast is called (after T+13.5sec) and will interrupt the spell only at that moment because target is out of range. As a caster I would like to be notified earlier about my target already out of range and not having to wait 12 seconds to realize.
This issue opens up the possibility for the target to go out of range, then do something and come back in range and the spell would still be successful.
For the spell to be successful the target needs to remains within range of the caster for the whole duration of the cast. What this PR does is to subdivide the 12 seconds interval in smaller chunks and repeat CheckDuringCast several times so if the target gets out of range, my cast is more or less immediately interrupted.

The way the checks are performed during a cast can be better visualized by putting a log in CheckBeginCast, CheckAfterCast, CheckDuringCast, CheckEndCast. Without this PR one can see the big gap between the checks. With this PR we can see that CheckDuringCast is repeated multiple times.

Regarding your second point I’m not sure to understand everything. A cast time interval of 15sec is actually not split equally in three 5sec periods due to the server property which limits step1 and step3 by default to 1500 (therefore step2 = 12000).

NetDwarf:
Sorry for the delay :disappointed: 

I could now reproduce the error. Very nice find! 

However the solution is hard to accept, since it is altering the signature of `SpellHandler.CheckDuringCast` so it's going to break many script that implements new SpellHandler. Furthermore the new argument is never used in the actual method.

Do you see a way to add this without changing method signature(s)? For example just changing the counter directly instead of in the method call?

Lotendan:
I have amended the commit :) Initial intention was to provide the subset parameter so the developer can act on the substep number in CheckDuringCast, but this also works without.

NetDwarf:
Thank you! :smiley:

PR #374, opened 2021-12-13
Use nameof to target table columns

Lotendan:
To make code refactoring much easier

### DIFF
--- a/GameServer/quests/QuestsMgr/QuestMgr.cs
+++ b/GameServer/quests/QuestsMgr/QuestMgr.cs
@@ -143,7 +143,7 @@ namespace DOL.GS.Quests
 				string tempID = Convert.ToString(identifier);
 
                 // TODO: Dirty Hack this should be done better
-                var mob = DOLDB<Mob>.SelectObject(DB.Column("Mob_ID").IsEqualTo(tempID).Or(DB.Column("Name").IsEqualTo(tempID)));
+                var mob = DOLDB<Mob>.SelectObject(DB.Column("Mob_ID").IsEqualTo(tempID).Or(DB.Column(nameof(Mob.Name)).IsEqualTo(tempID)));
 
                 GameNPC[] livings = WorldMgr.GetNPCsByName(mob.Name,(eRealm) mob.Realm);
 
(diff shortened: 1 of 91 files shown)


GameServer/quests/QuestsMgr/QuestMgr.cs
```diff
@@ -143,7 +143,7 @@ public static GameLiving ResolveLiving(object identifier, GameLiving defaultLivi
 				string tempID = Convert.ToString(identifier);
 
                 // TODO: Dirty Hack this should be done better
-                var mob = DOLDB<Mob>.SelectObject(DB.Column("Mob_ID").IsEqualTo(tempID).Or(DB.Column("Name").IsEqualTo(tempID)));
+                var mob = DOLDB<Mob>.SelectObject(DB.Column("Mob_ID").IsEqualTo(tempID).Or(DB.Column(nameof(Mob.Name)).IsEqualTo(tempID)));
```
Lotendan, line 146: I was not able to find a nameof for the table's auto-generated primary key. I think this is acceptable however. This kind of record should be retrieved from `FindObjectByKey` instead.

GameServerScripts/dbupdater/GuildAndAllianceUpdate.cs
```diff
@@ -42,12 +42,12 @@ public void Update()
 				log.Info("Start Searching for records that need update...");
 			
 			// Change the Leader Relation if Missing
-			var alliances = DOLDB<DBAlliance>.SelectObjects(DB.Column("LeaderGuildID").IsEqualTo(string.Empty).Or(DB.Column("LeaderGuildID").IsNull()));
+			var alliances = DOLDB<DBAlliance>.SelectObjects(DB.Column(nameof(DBAlliance.LeaderGuildID)).IsEqualTo(string.Empty).Or(DB.Column(nameof(DBAlliance.LeaderGuildID)).IsNull()));
```
NetDwarf, line 45: The usage of `nameof` in GameServerScripts does not work because it is using an older compiler unfortunately. The same applies for L50 in the same file of course.

Lotendan: Thanks, I have updated the PR and excluded GameServerScripts from the use of nameof.
Additionally I have created another commit to change the names of the relation columns in the DB.
Still I miss the _ID column name but that would require to have it computed at compile time.

NetDwarf:
Thank you! It's merged.

PR #378, opened 2022-02-06
Replace deprecated classes for Net6

NetDwarf:
Replace RNGCryptoServiceProvider with RandomNumberGenerator
Replace MD5CryptoServiceProvider with MD5
Replace SHA256Managed with SHA256
Replace WebRequest/WebClient with HttpClient
Add reference to System.Net.Http to DOLScriptCompilers


NetDwarf:
It would be nice if someone could test the new ServerListUpdate.cs. Maybe @DigitalBox98 ?

DigitalBox98:
@NetDwarf I can check that, but I will need before that to clone your repo in the maintenance branch to make the test :)

PR #380, opened 2022-03-16
Load Mobs In disabled Zones server property

Loki1976:
If we are disabling zones it would seem we should disable the loading of mobs in these zones also to free up some resources. So added a server property to allow feature.

### DIFF
--- a/GameServer/serverproperty/ServerProperties.cs
+++ b/GameServer/serverproperty/ServerProperties.cs
@@ -606,6 +606,12 @@ namespace DOL.GS.ServerProperties
 		public static int SET_DIFFICULTY_ON_EPIC_ENCOUNTERS;
 
 		/// <summary>
+		/// Load mobs in disabled zones
+		/// </summary>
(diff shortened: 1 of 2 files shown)


NetDwarf:
Nice catch and thank you for your contribution! 

Is there a scenario where you would want to load the mobs in disabled zones? Otherwise I would consider the previous behavior a bug, so we could actually do without the ServerProperty.

Loki1976:
Hi, I'm not aware of any reason why this server property exists. I  just made the changes to a server I am working with and made the contribution back to the project. I would agree that we could  loose this property also.

NetDwarf:
Could you remove the ServerProperty then and make the new behavior (not loading mobs in disabled zones) default? :slightly_smiling_face:

Loki1976:
Yeah sure. I will close this PR and make changes with a new PR.

NetDwarf:
You can also force-push a change. Admittedly `git` has a bit of a learning curve to it, but it's a nifty tool (especially CLI).

PR #381, opened 2022-03-23
fix packet: writing pascal string off by one + max length

JVirant:
A fix with big changes, it fixes the character overview for 1.125+.
The WriteByte(0) seems odd but it was what I got from official server (with 1.127e).

After a small time with disassembled game.dll, I retrieved some string limits.

### DIFF
--- a/DOLBase/Network/PacketOut.cs
+++ b/DOLBase/Network/PacketOut.cs
@@ -202,7 +202,7 @@ namespace DOL.Network
 			Write(bytes, 0, bytes.Length);
 		}
 
-		public void WritePascalStringIntLE(string str)
+		public void WritePascalStringIntLE(string str, int maxlen = 2048)
 		{
 			if (str == null || str.Length <= 0)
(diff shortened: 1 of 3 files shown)


JVirant:
Actually, it's not truncated (at least for 1.126), from my disassembly skill I understand something like that:
`int readBuffer(Packet* packet, char* buffer, int bufferLen)
{
  // read the int32 le from pascal string
  int len = (int)(packet->data + packet->position);
  packet->position += 4;
  if (len > bufferLen)
    // return with an error
  if (packet->position + len > packet->length)
    // return with an error
  // read the buffer
}

// just for info:
struct Packet {
  char* data;
  int length;
  int position;
}
`

The 0x18 value was just from calls to this function, I may be in error but it should be fine for most versions and it's used only in few packets (CharacterOverview, MerchantWindow Regions) or I missed some.

I found this issue with the location description, not with the character name (it's mostly ok for this part :)).

PR #383, opened 2022-04-05
Do not delve the 'weapon' stats of an instrument

Lotendan:
An instrument does not have DPS or SPD, Effective Damage, etc...

For instruments, DPS_AF is used to indicate what kind of instrument the object is (drum, lute, flute, harp).
Currently, delving instruments show "0.1 Base DPS / 0.1 Clamped DPS / 0.1 Effective DPS" etc...

With this change I remove `DelveWeaponStats` which is not necessary for instruments.


NetDwarf:
Just recently I looked at instrument details and didn't even notice. Nice catch!

Thank you! :+1:

PR #386, opened 2022-05-26
Adjustment on keep lord auto-heal for low level BG

DigitalBox98:
Possibility to remove the Keep Lord to heal itself 
0=no self heal
1=self heal (which is the default)
On low level keeps (battlegrounds), the keep lord seems impossible to kill with a little group if he has a non-interruptible self heal

### DIFF
--- a/GameServer/ai/brain/Guards/Lord.cs
+++ b/GameServer/ai/brain/Guards/Lord.cs
@@ -15,26 +15,23 @@ namespace DOL.AI.Brain
 
 		public override void Think()
 		{
-			if (GS.ServerProperties.Properties.KEEP_LORD_HEAL_ITSELF == true)
-            {
-				if (Body != null && Body.Spells.Count == 0)
+			// Add auto heal for lord above level 15
+			if (Body != null && Body.Spells.Count == 0 && Body.Level>=15)
+			{
+				switch (Body.Realm)
 				{
-					switch (Body.Realm)
-					{
-						case eRealm.None:
-						case eRealm.Albion:
-							Body.Spells.Add(GuardSpellDB.AlbLordHealSpell);
-							break;
-						case eRealm.Midgard:
-							Body.Spells.Add(GuardSpellDB.MidLordHealSpell);
-							break;
-						case eRealm.Hibernia:
-							Body.Spells.Add(GuardSpellDB.HibLordHealSpell);
-							break;
-					}
+					case eRealm.None:
+					case eRealm.Albion:
+						Body.Spells.Add(GuardSpellDB.AlbLordHealSpell);
+						break;
+					case eRealm.Midgard:
+						Body.Spells.Add(GuardSpellDB.MidLordHealSpell);
+						break;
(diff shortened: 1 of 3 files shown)


NetDwarf:

This would remove healing for all lords, right? That would be quite a niche use case and therefore not warrant a custom ServerProperty I think, but the current implementation is indeed an issue. At first I thought this could be fixed with the ServerProperty controlling the threshold, however this is bundled with other NPCs using the same one (it's `npc_heal_threshold` switching that to `keep_heal_threshold` would make this more feasible I think). 

So one could add a ServerProperty for controlling the threshold of GuardLords, but that would still not solve the root problem, that is the static and unremovable heal.

To solve this one could for example make the heals relative to Lord level for example make the heal value 4*Level. That would not remove the heal, but would make the lord somewhat defeatable at any given level. That would still leave the heal unremovable from the Lord, tho. 

Another approach would be to make NPCTemplates for specific lords, but this comes with its own set of drawbacks. But it would be the most customizable and also mostly sane if one can find out how to assign them to the given Lord/Lady.

NetDwarf:
Actually, using a percentage value would be easier for scaling the heal and would be a simple fix. Would that also fit your use case? (even low lords should be doable like that)

NetDwarf:
> I like the idea of using a percentage, so if I understand it would be : spell.Value = 4 * m_level

Even better you can use % values. In code this would be a negative value (HealSpell.Value = -3 means 3% heal). Something like 2.5% (-2.5) seems adequate.

> We could update this approach by having the spell value very low (or 0) below a certain level of Lord ?
Like if level of lord under 40 (to be determined for the battlegrounds or by a server property), we would apply this mechanism.

I would say that only the really low lords should be considered with no heal at all. Maybe until L15 or so. And not by changing the heal value, but by removing the spell cast I'd say.

DigitalBox98:
Ok I've checked with a percentage of 2.5% and it still quite difficult at low level.
On live on low level BGs, I didn't notice any heal from Lord but I don't know at which level or BG it appears.

So the best would be to have low lords with no heal at heal (until which BG?). So the check of the level of lord to authorize the auto heal would be the best approach : )

One possibility is to have the threshold value defaulted to a specific value (15? 20? 30?) and to authorize to overwrite it via a server property?

NetDwarf:
> Ok I've checked with a percentage of 2.5% and it still quite difficult at low level.

How difficult? I mean alone against the lord should be difficult for example :slightly_smiling_face: 

> One possibility is to have the threshold value defaulted to a specific value (15? 20? 30?) and to authorize to overwrite it via a server property?

Hardcode it. Nobody is gonna use this property anyway. I'd say above L15 the "natural" health regen should be low enough to not play a major role anymore. And to fix this in general we need a different approach, but we need a bandaid in the meantime (read: maybe forever).

DigitalBox98:
Ok the property has been removed and the auto heal is now a percentage of the health of the lord + auto heal not activated below lord level 15

NetDwarf:
Looks fine. Thank you!

PR #389, opened 2022-06-07
Use a static dictionary to hold the locks for all logins

Lotendan:
The dictionary is currently a member of the LoginRequestHandler class, which means that there is one Dictionary per account, therefore there is only ever one item added to it and one lock held! This defeats the purpose of having a dictionary at all for the login process.
Changing the dictionary as static instead makes it relevant to actually acquire server-wide locks.


NetDwarf:
Let's merge it! In theory this should fix a missing lock as LoginRequestHandler is member of GameClient (via PacketProcessor) which is not bound to an account, so re-entry is not actually prevented by the old implementation.

PR #391, opened 2022-06-15
Battleground : addition of GameKeepBanner in the battleground keeps

DigitalBox98:
In case of usage of the /keep addbanner realm command, the GameKeepBanner are now taken into account for the battleground keeps :)


NetDwarf:
Sorry for the delay.

Banners in BGs now show up correctly.

Thank you!

PR #392, opened 2022-06-21
Revert "Update Microsoft.Data.Sqlite to 6.0.5"

NetDwarf:
This reverts commit 93a3d8ba766983d5f8d7e765f4601413cacb8afe.
This fixes startup on Windows with SQLite DB


NetDwarf:
On Windows the broken version (2.0.6) of `SqlitePCLRaw.core` is enforced as dependency of `Microsoft.Data.Sqlite` for reasons unknown to me. Preview of version 7 of `Microsoft.Data.Sqlite` uses a newer `SqlitePCLRaw.core` version and at least in preliminary testing worked, so we probably have an upgrade path later.

PR #394, opened 2022-06-22
SiegeRam : new InventoryItem class to create a siege ram

DigitalBox98:
This new class will require 4 new itemtemplates in the DB corresponding to each kind of siege ram and must be linked to DOL.GS.GameInventorySiegeRam via the ClassType field

This PR will be linked to another PR on the db-public repo to complete this enhancement


NetDwarf:
Thank you! It works exactly like advertised.

PR #399, opened 2022-08-17
Add RealmEnhancementMerchant CustomNPC

DigitalBox98:
RealmEnhancementMerchant will propose (live-like) a buff token for level 10+.
This token can be paid with gold or bounty points and will apply 7 buffs to the player

Additional tokens can be added on a later stage for further levels/BG


NetDwarf:
Thank you for your contribution! (and sorry for the late reply)

This has some resemblance to `DOL.GS.BuffMerchant`. I tested them both the same way, but `DOL.GS.RealmEnhancementMerchant` has two items in the merchant list, one for buff token for 32 gold and one for 32 copper (which is probably meant for BP). I also couldn't switch to BP (same problem I think). 

You should be able to setup a scripted merchant list with `MerchantCatalog`. You don't even need the dialog, just put them on two different pages and set a different currency for each page.

In general I feel like most of the data belongs in the database and the buffer just consists of the merchant profession and the buffer profession, however it would need some rework, which one needs time for. :sweat_smile:

DigitalBox98:
The RealmEnhancementMerchant behavior is fine with the latest db-public database, but I will look at the case where the spells/tokens are not in the DB (I think it's the case you mention with the issue on the BP switch)

"Putting the tokens on two different pages without any dialog and set a different currency for each page" : it was my initial idea, but as I mimick DOL.GS.BuffMerchant, I keep the dialog part. I will review that without the dialog part.
It should be more easier to use but I need to understand how to do that.

"In general I feel like most of the data belongs in the database and the buffer just consists of the merchant profession and the buffer profession" : what do you mean ? 
I was thinking on relying only on the database for the spells/tokens data, but as DOL.GS.BuffMerchant was mixing both (database or spell/token/... creation in case they don't exist in DB), I keep also this behavior. 
Relying only on the database could be an easier approach maybe.

NetDwarf:
> The RealmEnhancementMerchant behavior is fine with the latest db-public database, but I will look at the case where the spells/tokens are not in the DB (I think it's the case you mention with the issue on the BP switch)

That's a bit of a problem, since it's hard to find out what data is needed for this to work I think. 

> Relying only on the database could be an easier approach maybe.

Likely not easier, but cleaner, more composable and less redundant. :slightly_smiling_face: At any rate. Creating a MerchantCatalog should be half-way there :)

Another idea is to only keep the buffer ability (buff on receiving a token), inherit from `GameMerchant` and handle the merchant list as usual. (see https://github.com/Dawn-of-Light/DOLSharp/releases/tag/1.9.7.3852 for multi-currency)

DigitalBox98:
This new version is greatly simplified and is only relying on DB data :)
The MerchantItem is handling both BP and gold token versions up to the slot -1 for the currency in the DB.

NetDwarf:
Sorry for the late reply (again :weary: ). Moving the MerchantItemList out seems fine, however hardcoding the spell IDs is not nice. I have an idea or actually two how to solve it (even if it is a bit hacky), so am going to append that in the next few days :slightly_smiling_face:

DigitalBox98:
Ah yes, the spellID is the way it's working today. 
Having a field equivalent to "Id_nb" from ItemTemplate applied to Spell table would be better to identify the spells but not sure that this kind of change is simple to do.
So I'm curious about your "hack" ;)

NetDwarf:
Disclaimer: I haven't implemented anything yet. (Maybe a good thing)

I don't like that anything is hardcoded there. The basic idea of my hack would be to add the SpellID in the item itself. The item is part of the MerchantItem(List) and the Merchant can determine which items it accepts from that list. 

Now, one SpellID would not exactly be enough and there are (at least) two ways to implement that. Either a spellcombo (multiple subspells) or string list of spells (which I don't like much). Not sure how much needs to be changed (if anything at all outside of the Merchant class).

NetDwarf:
You can test it by adding this to your DB:
```
UPDATE ItemTemplate 
SET Effect = 225823756
where Id_nb IN ('BG_BP_Realm_Full_Buffs_Token_10', 'BG_Realm_Full_Buffs_Token_10');

INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823757', 44, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823758', 45, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823759', 46, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823760', 47, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823761', 48, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823762', 49, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823763', 50, '2000-01-01 00:00:00');
INSERT INTO SpellXCustomValues
(SpellID, KeyName, Value, CustomParamID, LastTimeRowUpdated)
VALUES(225823756, 'MultipleSubSpellID', '225823764', 51, '2000-01-01 00:00:00');
```

And then creating a `DOL.GS.SpellTokenMerchant`

DigitalBox98:
Ok I have checked it and it's working perfectly fine :)

PR #406, opened 2022-10-01
Force respawn MissionMaster on Tower/Keep take

NetDwarf:


NetDwarf:
It feels like we should move the initialization tasks to `GameLiving.AddToWorld`, but I am not ready to deal with the impact, since I am not sure of the edge cases here. Or viewed differently: I am not sure why `AddToWorld` does not do mostly the same tasks as the `RespawnTimerCallback`.

This fix is btw a small SNAFU after changing respawn timer of MissionMaster. The intended behavior is of course that it spawns instantly after keep/tower take, but if it is killed by a player it should take at least some time to respawn to avoid getting the player stuck in a fighting loop against the MissionMaster :smile:

PR #409, opened 2022-10-13
Remove BP cap

NetDwarf:


NetDwarf:
BP is now a currency and therefore the cap does not seem adequate anymore. The cap also collides with other values, especially GuardLord BPs. Furthermore is it not associated with a progress directly like CLXP, XP or RP are.

PR #410, opened 2022-10-16
Fix casting GamePets

NetDwarf:
Make spells available to GamePet
Fix Bonedancer hybrid subpets when commander is a caster


NetDwarf:
Okay, this was really painful. It started with Bonedancer healers not working and turned out to be GamePets not being able to cast at all. This is possibly imperfect as `InCombat` has some limitations, but pets should now be able to cast their respective spells.

PR #411, opened 2022-10-16
Remove pet buffing from BuffMerchant

NetDwarf:


NetDwarf:
This is not the case on Live and as far as I can tell never was.

PR #414, opened 2022-10-27
Fix aggro management on stun/mezz

NetDwarf:
Set initial aggro of 'added' mobs to 1

### DIFF
--- a/GameServer/ai/brain/StandardMobBrain.cs
+++ b/GameServer/ai/brain/StandardMobBrain.cs
@@ -232,8 +232,7 @@ namespace DOL.AI.Brain
 		/// </summary>
 		protected virtual void CheckNPCAggro()
 		{
-			if (Body.AttackState)
-				return;
+			if (HasAggro) return;
 
 			foreach (GameNPC npc in Body.GetNPCsInRadius((ushort)AggroRange, Body.CurrentRegion.IsDungeon ? false : true))
 			{
@@ -260,9 +259,7 @@ namespace DOL.AI.Brain
 		/// </summary>
 		protected virtual void CheckPlayerAggro()
 		{
-			//Check if we are already attacking, return if yes
-			if (Body.AttackState)
-				return;
+			if (HasAggro) return;
 
 			foreach (GamePlayer player in Body.GetPlayersInRadius((ushort)AggroRange, Body.CurrentZone.IsDungeon ? false : true))
 			{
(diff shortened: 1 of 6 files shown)


Lotendan:
Hello hello, can you please comment on the reasons this helps with on aggro management for stun/mezz?

GameServer/ai/brain/StandardMobBrain.cs
```diff
@@ -373,17 +370,7 @@ public virtual int AggroRange
...
-			{
-				bool hasAggro = false;
-				lock ((m_aggroTable as ICollection).SyncRoot)
-				{
-					hasAggro = (m_aggroTable.Count > 0);
-				}
-				return hasAggro;
-			}
-		}
+			=> AggroTable.Count > 0;
```
Lotendan, line 373: You are removing a lock here. This gets dangerous for concurrent access.

NetDwarf: Access to `.Count` is thread-safe as it is calling members of `Dictionary`. Generally speaking `.Count` is meant to be thread-safe/atomic while `.Count()` does not have this promise. Granted the value might mismatch after it has been called, but the lock is not protecting against that.

NetDwarf:
> Hello hello, can you please comment on the reasons this helps with on aggro management for stun/mezz?

`CheckPlayerAggro` seems to be used for when you come too close to mobs (not in combat) and they aggro on you (aka "add") and added 100 aggro previously. This was "aborted" when `GameLiving.AttackState` was set to `false`, however this is also the case when someone is mezzed or stunned and adds 100 aggro (for L50) on every `Think()` tick for everyone in radius except the current target.

The initial aggro was also reduced to 1 in the process as this is very light aggro on Live as well.

Lotendan:
This makes a lot of sense, thank you!
Is there any reason why the following locations still have +1 aggro or is this a miss?
https://github.com/Dawn-of-Light/DOLSharp/blob/master/GameServer/ai/brain/Animist/TurretFNFBrain.cs#L140 
https://github.com/Dawn-of-Light/DOLSharp/blob/master/GameServer/ai/brain/Animist/TurretMainPetCasterBrain.cs#L163
https://github.com/Dawn-of-Light/DOLSharp/blob/master/GameServer/ai/brain/Animist/TurretMainPetTankBrain.cs#L79
(maybe there are more, haven't checked)

NetDwarf:
> Is there any reason why the following locations still have +1 aggro or is this a miss?

Mostly it's a tradeoff. I reduced the value drastically, but didn't want to risk breaking behavior that expects a non-zero value.

PR #423, opened 2022-11-26
Update Github Actions

NetDwarf:
Replace deprecated set-output with GITHUB_ENV


NetDwarf:
Updated according to:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #424, opened 2022-11-29
moved hard coded text from playercommands into language file.

Mavcraft:
I have removed the hard coded text and added all english translations into PlayerCommands.txt.


NetDwarf:
Looks good to me! Thank you!

PR #425, opened 2022-12-03
Almost all playercommand sentences moved to language file

Mavcraft:
Here are a lot more sentences from commands moved to language file.
My studio had an issue with line breaks in webdisplay.cs, that's why so much lines are marked.


NetDwarf:
Thank you! :slightly_smiling_face:

PR #427, opened 2022-12-13
KeepSkin : use new keep skin by default

DigitalBox98:
As the old keep skin is giving some blank texture, proposal is to use the new keep skin by default which is better supported.


NetDwarf:
As far as I can tell this is still broken. The old default seems therefore better until we manage to move guards/banners/hastener etc to new keep layout.

PR #428, opened 2022-12-15
KeepDoor : transversal keep door need a larger size

DigitalBox98:
As the side doors have some "air" space between external and internal part, the size of the jump need to be extended to 175


NetDwarf:
Thank you!

PR #433, opened 2023-01-03
Accept currency item if ID is the same

NetDwarf:
Merchants accept also ItemTemplates that have no ClassType but same Id_nb as ItemCurrency.id
Separate ToText() and ID
Determine ToText() by ItemTemplate.Name of Item.Id_nb == ItemCurrency.id 
Fix wrong display of Copper MerchantCatalogPages that follow other ones

### DIFF
--- a/GameServer/Finance/Currency.cs
+++ b/GameServer/Finance/Currency.cs
@@ -1,10 +1,10 @@
 using System;
+using DOL.Database;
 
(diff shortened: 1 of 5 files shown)


tegstewart:
Definitely fixes the problems I was just having with DF merchants not working.  :-)

PR #436, opened 2023-01-11
Update CharmSpellHandler.cs

tegstewart:
Made error messages when charms fail more useful.

I found the switch checking body and charm types with gotos and multiple if statements really unintuitive, so I redid it in a way I think is much more legible.


NetDwarf:
LGTM! Thank you!

PR #437, opened 2023-01-12
Auto Level CLs

tegstewart:
Are you tired of schlepping your mailed heinie back to Jordheim, just to get a patronizing pat on the head from the king?  Is that throne room access guard in TnN undressing you with his eyes every single time you walk by?  Is running around Camelot, ON FOOT, simply beneath your station?

Set the AUTO_CHAMPION_LEVELUP server property to true, and leave all that behind today!

Disabled by default, because it's not livelike.


NetDwarf:
Thank you for your contribution! :slightly_smiling_face: 

I think we can do without the ServerProperty. You can make it the default. This is Live-like anyway.

tegstewart:
Will do.

PR #438, opened 2023-01-12
Auto Level CLs

tegstewart:
Automatically increase CLs when appropriate, rather than having to go to town to talk to the king.


NetDwarf:
Thank you! :tada:

PR #439, opened 2023-01-15
Added Eligible Races

tegstewart:
Added missing eligible races, according  to 1.127e

### DIFF
--- a/GameServer/playerclasses/albion/ClassFriar.cs
+++ b/GameServer/playerclasses/albion/ClassFriar.cs
@@ -45,7 +45,7 @@ namespace DOL.GS.PlayerClass
 
 		public override List<PlayerRace> EligibleRaces => new List<PlayerRace>()
 		{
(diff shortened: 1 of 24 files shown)


NetDwarf:
Hm, I am not able to find patch notes regarding this. Could you give me a pointer? :sweat_smile:

tegstewart:
No patch notes, I went through the client and entered the combinations the client allows that DOL doesn't.

tegstewart:
1.125 added six combos based on forum polls, which is kinda neat:

Highlander Infiltrator, Saracen Heretic, Dwarf Shadowblade, Valkyn Valkyrie,
 Sylvan Ranger, Graoch Champion


I missed the last one, as the client isn't letting me select minotaurs at all, maybe an EC thing?


I wonder if it would be better to move eligible races .  Maybe a DB table, and add a player command to display and switch to available races to bypass client restrictions entirely.

NetDwarf:
I'd be in favor of only adding the ones mentioned in 1.125 for now. Having that in DB (customizable) would be indeed sensible.

PR #443, opened 2023-02-02
Improved Stealth for NPCs

tegstewart:
This is relatively simple, but important, and opens up a lot of future possibilities.  Under the previous implementation, NPCs with the Stealth flag were always stealthed, so logic checking IsStealthed always returned true.  This lead to GameLiving.IsAttackable always returning false, and CheckNPCAggro() scrubbing NPCs with the Stealth flag.  Stealthed NPCs could attack and kill other NPCs with impunity, as the other NPCs had no mechanism to detect them, and essentially just ignored them while being murdered.

I've added a CheckStealth() method which is called at the end of StandardMobBrain.Think().  NPCs remain stealthed until they enter combat, and unstealth once they're out of combat.  When Flags is updated, it sends an update packet to the client, so NPCs stealth and unstealth both visually and logically.

I also added accessors for the other eFlag types, mostly because we have manual bit masks all over the place, I was getting annoyed retyping them out every time, and accessors are so much more readible.  I considered adding set accessors as well, but I tend to think if somebody is fiddling with flags, they really need to know how they work, and could create other methods like Stealth() to do it safely.


GameServer/gameobjects/GameNPC.cs
```diff
@@ -756,12 +760,58 @@ public virtual eFlags Flags
 			}
 		}
 
+		public bool IsGhost
+		{ get => (m_flags & eFlags.GHOST) != 0; }
+
+		public override bool IsStealthed
+		{ get => (m_flags & eFlags.STEALTH) != 0; }
```
Lotendan, line 767: `m_flags.HasFlag( eFlags.STEALTH )` ?

tegstewart: Hmm.  It might be better to use HasFlag than the accessors.

I'm happy either way.

tegstewart:
This should be ready to go.  I've beat my head against a wall testing it.

NetDwarf:
Thank you for your work! Works like a charm! :partying_face:

PR #444, opened 2023-02-03
Move CharacterClass to DB

NetDwarf:
Refactor CharacterClass
Remove ICharacterClass and all its implementations
Save default CharacterClasses into DB if they do not exist yet
Fix pet window to no longer close when releasing subpet while having no main pet
Change Necromancer spells only using power when pet starts casting them


NetDwarf:
As usual the "last minute" change broke the tests and I should've seen it coming.

For anyone wondering why a change to CharacterClass would change anything related to pets should take a look at [ICharacterClass](https://github.com/Dawn-of-Light/DOLSharp/blob/692b3effb31b2577d83b73a994db0d804296ddc5/GameServer/gameobjects/CharacterClasses/ICharacterClass.cs). There is also a lot of room for bugs, so feel free to test it.

PR #448, opened 2023-02-09
Remove Net Framework

NetDwarf:
Add option to load libraries from lib folder
Move libraries in release build to lib folder
Remove 32-bit, WASM, S390X, mips64, alpine,maccatalyst SQLite binaries from release
Replace AppHost binary with script (batch and shell)
Adjust Docker builds for NetCore only


NetDwarf:
You can test the final builds here: https://github.com/NetDwarf/DOLSharp/releases/tag/1.9.7.3717

I cannot yet test docker builds, so that might be a sore spot. It would be nice if @dol-leodagan would look over these changes.

NetDwarf:
Let's do it!

Issue #451, opened 2023-02-17
Warlock chambers displayed incorrectly


xGohac:
When the warlord has used up his chambers, an orb will appear over the character when it shouldn't be there anymore.

NetDwarf:
Thank you for your submission!

I currently try to replicate it, but I am unable to. Which client version and spell order did you use?

PR #452, opened 2023-02-20
Fix Warlock Chamber Display

NetDwarf:
Previously there was one too many chamber displayed
Add empty chamber if spell has no corresponding model 
Remove chamber effect packet spam


NetDwarf:
This fixes #451 .

PR #460, opened 2023-06-14
Path finding

jeremv42:
It add path finding based on Uthgard sources, I fixed some threading issues and added some missing code.

For now, the code use path finding only for NPC roaming, I think we need more returns/tests to see if using it everywhere is OK. I did it on roaming because it can fix one of the worst issue: mob traversing wall in dungeons.

To build navigation meshes, you need to use Uthgard tools (https://github.com/thekroko/uthgard-opensource/tree/master/pathing/buildnav) and put the nav files in a folder "pathing" where you have your DOLServer.dll.
If you don't want to take time to build every navigation meshes, I generated many of them a few time ago: https://amtenael.fr/pathing.7z

dol_detour.dll is the equivalent of closed source ReUth.dll from Uthgard, I automated the build for AMD64 in Github actions, you can follow some instructions in Pathing/Detour/README.md.

Caution: if you load all nav meshes, you will need at least 5GB of ram (maybe more, I didn't check).


NetDwarf:
That looks exciting!

NetDwarf:
Finally had the time to test it.

It's a bit tricky to get working in Linux/Flatpak (without gcc, had to use `libltdl.so`), but it seems to work in roaming (as implied in your initial post it doesn't work for attacks). Great job!

For the structure and how it's introduced I am not quite sure yet. Having libraries compile for each version seems suboptimal in the release process. Possibly it could reside either in another project inside DOL or another repo in Github. The latter is maybe bad for discoverability, but in that case it makes more sense to compile for different platforms/archs. We could give it the same treatment as DOLConfig then maybe.

But maybe I put the cart before the horse. There are some questions:

1. Is that the final version for now or is it in a drafting phase of sorts?
2. Could we move to managed eventually? (removing the necessity to compile for different platforms)
3. How resilient is it in terms of CPU usage? (Can pathing cause DoS?)

At any rate, pathing would be such a boon for DOL.

PS.: For which client/model version did you extract the navmesh?

jeremv42:
> * Is that the final version for now or is it in a drafting phase of sorts?

For the C++ lib, it's ok. Regarding Dol changes, as I said, I only implemented the roaming part, it needs many changes for following and other npc movements.
If you want to separate it from .net, I think we need to add a version somewhere. Having an incorrect call to native code will lead to segmentation fault, so it's better to know incompatibilities before calling any C function.

> * Could we move to managed eventually? (removing the necessity to compile for different platforms)

I don't think it's possible. Maybe you can build the C++ as C++ CLI but I never used this language, I'm not sure you can compile it in another platform than windows...
You can find some people who rewrote the Detour lib in C# but I prefer to use the original one.

> * How resilient is it in terms of CPU usage?

I can't say if it's too much or not, but I checked the usage of Detour in other games (especially in AzerothCore, disclaimer: I never managed or used AzerothCore, just read some code) and it seems to be used in the same way: call the functions without any scheduling (no threading, no async,...).
I think the biggest issue is more about RAM usage: with all my navmeshes, it took ~7GB of RAM so it's nearly 10 times more (and it's not displayed in managed memory usage so the stat printer is wrong).

> Can pathing cause DoS?

I can't say anything about security (DoS), I didn't write the Detour lib. It's a path finding algorithm a bit like A* with a limit for the number of nodes for a path so I don't think it can be used as denial of service.
About threading, the Uthgard version had a big issue: it doesn't take threading safety into account. Detour is not thread safe but you can be safe by separating dtNavMesh and dtNavMeshQuery usages.
dtNavMesh holds the navigation mesh which is static, so we can threat it as read-only. dtNavMeshQuery holds the data to do the path finding which can't be shared among threads. If we want to manage dynamic blocking objects (like keeps), we will need to carefully manage locks to edit the dtNavMesh.
My implementation uses thread locals to manage dtNavMeshQuery with disposable instances to free the C++ instance when a thread exits (see NavMeshQuery and _navmesQueries in LocalPathingMgr.cs).

> PS.: For which client/model version did you extract the navmesh?

I don't remember, it's probably something like 1.126, it's a bit time consuming to generate all navmeshes especially with the old version of Recast from Uthgard tools.
One day, I will try to redevelop this tool so we will be able to update it (newer Recast version are faster). It will be useful to add it to Dol too to manage dynamic changes: keeps, some blocking objects, etc.

NetDwarf:
Thanks for the answers! 

> For the C++ lib, it's ok. Regarding Dol changes, as I said, I only implemented the roaming part, it needs many changes for following and other npc movements.

It's not done for sure, but it's a good starting point. If there are no navmeshes added it shouldn't change behavior on the user side at all (which is important and good). The licensing should also not be a problem (from what I've seen zlib license is compatible with GPL2).

> If you want to separate it from .net, I think we need to add a version somewhere. Having an incorrect call to native code will lead to segmentation fault, so it's better to know incompatibilities before calling any C function.

I guess for now we keep it the way it is. Because the options are mutually exclusive, so we decide/act then.

I'd merge it, if you're okay with that. :slightly_smiling_face:

PR #464, opened 2023-07-28
Enable NavMesh Pathing for Following/Attacking

NetDwarf:
Also add pathing for Goto, Here and several other commands


NetDwarf:
From limited testing it seems to work mostly as expected. Of course mobs sometimes run too close to the edge and then appear on a level below just to run on a level above again. But pretty minor issues actually.

Is there a scenario missing, @jeremv42 ? I only tested (all with at least one obstacle):
1. Directly casting on a mob
2. Send pet to attack another mob
3. Set pet on follow

PR #470, opened 2023-09-14
Update WorldMgr.cs

Loki1976:
Prevent loading mobs in disabled regions for server performance.


Loki1976:
Ok, firing up my old laptop and doing a pull request was a bad idea. Let me come back to this.

PR #472, opened 2023-09-20
Refactor Point3D/GameLocation

NetDwarf:


GameServer/packets/Client/168/PlayerPositionUpdateHandler.cs
```diff
@@ -313,25 +309,16 @@ public void HandlePacket(GameClient client, GSPacketIn packet)
...
-			client.Player.Heading = (ushort)(headingflag & 0xFFF);
-			if (client.Player.X != realX || client.Player.Y != realY)
+			if (client.Player.Location.X != newLocation.X || client.Player.Location.Y != newLocation.Y)
+            {
 				client.Player.TempProperties.setProperty(LASTMOVEMENTTICK, client.Player.CurrentRegion.Time);
-			client.Player.X = realX;
-			client.Player.Y = realY;
-			client.Player.Z = realZ;
+            }
+            client.Player.Position = Position.Create(newZone.ZoneRegion.ID, coordinate: newLocation, heading: (ushort)(headingflag & 0xFFF));
```
jeremv42, line 316: I think, you can have a data race condition, C# says an assignment to an int (for 32bits, a long for 64bits) is atomic but not for a struct.
I did the same on my server and sometimes, I had an invalid position in other parts of code but I continue to have some even with an Interlocked call.

GameServer/Geometry/Vector.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+
+namespace DOL.GS.Geometry;
+
+public struct Vector
```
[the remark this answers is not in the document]
NetDwarf: One can change the internal implementation of the Vector class later on. As we need integers at some point I feel like using int upfront is maybe the better call, but I'd be surprised if it is noticable at all either way.

GameServer/packets/Client/168/PlayerPositionUpdateHandler.cs
NetDwarf: How/Where did you have invalid positions? This sounds more like a bug rather than a race condition. I could make Position and Coordinate a class of course, but not sure if it is actually possible to cause a race condition like this.

NetDwarf: I did some testing and could trigger a race condition with struct assignments (regardless of the byte length btw). This is fixable by making `Position` a class (as expected), however given that the `GameLiving.Position` is inside an immutable class (`Motion`) this is already only a problem in static objects (the ones that are not subclasses of GameLiving), where the `Position` is rarely changed.

It would be valuable to know where/how you had invalid positions, as this is likely a bug still.

jeremv42: Yes, it's fine with a class because a reference assignment is atomic in C#.

I don't think it will be very useful, my server is heavily modified and I use Vector3 which use XMM registers internally (maybe it's even worse for data race conditions).

PR #477, opened 2023-10-02
Proposal for DataQuestReward quest

DigitalBox98:
The aim of this PR is to discuss about improvment to the current quest system. I don't know if both current DataQuest and new DataQuestRewardQuest must coexist or if one can replace the other.

With the new DataQuestReward quest we can have : 
- Each goal to reach can display a visible red dot on the map
- Each goal can have an item to take x times
- An icon displayed in the quest journal for each item retrieved 
- When a step is completed, a nice message is displayed to the player

quest1
quest2

The DataQuestReward quest is coming from OnTheDOL fork reintegrated in original DOL. The details of a quest example is here : http://losojos-001-site1.btempurl.com/Guides/Quests/doltoolboxrewardquests

To test it : 
- Create the 2 missing tables (dataquestrewardquest, characterxdqrewardq) or First launch the server and empty tables will be created 
- Insert a quest in the new table : https://github.com/Los-Ojos/Reward-DataQuests/blob/master/Albion/A%20Wizard's%20Aid(10).txt

Missing tables to create : sql files attached 

[table1.txt](https://github.com/Dawn-of-Light/DOLSharp/files/12786960/table1.txt)
[table2.txt](https://github.com/Dawn-of-Light/DOLSharp/files/12786961/table2.txt)


DigitalBox98:
They can coexist, however with this PR I'm currently studying if DataQuestRewardQuest is not the same than DataQuest with questType=RewardQuest but with enhancements. If so the idea would be to improve the current DataQuest :)

NetDwarf:
There are only two tasks open from my perspective:
1. Fix the warnings
2. Better table names (i.e. `RewardQuests` and `CharacterXRewardQuest`)

The latter is important to me because you can't easily change table names after the fact and the names are currently a bit overloaded and hard to decipher (`DataQuestRewardQuest` and `CharacterXDQRewardQ`). 

After this I'd merge it even if it contains bugs. We can still change them after the fact and the same applies to the merging of DataQuests and RewardQuests (if possible).

DigitalBox98:
Ok I will review that points :)

DigitalBox98:
Ok the warnings are now fixed and I have renamed the table to "RewardQuest" and "CharacterXRewardQuest"

PR #488, opened 2024-10-28
Allow NPC Protect Effect

tegstewart:
A while back, I gave custom pets the protect ability to apply to their owners, but it never worked properly as ProtectEffect assumed both the source and target of protect effects were players.  This remedies that by allowing the effect to be applied by and on GameLivings.

The ability handler already limits players to applying the effect to other players in their group, so vanilla behaviour hasn't changed, and the only other possible application currently is custom pets protecting their owners.  That said, I tried to future proof it so that other contexts are possible.  Somebody could create a group vs group fight between NPCs, and have tank NPCs protect healing NPCs, for example.

While I was there, I cleaned up the code and addressed compiler warnings and recommendations.

### DIFF
--- a/GameServer/ai/brain/ControlledNpcBrain.cs
+++ b/GameServer/ai/brain/ControlledNpcBrain.cs
@@ -464,7 +464,7 @@ namespace DOL.AI.Brain
 						case Abilities.Protect:
 							{
 								if (GetPlayerOwner() is GamePlayer player)
-									new ProtectEffect().Start(player);
+									new ProtectEffect().Start(Body, player);
 								break;
 							}
 						case Abilities.ChargeAbility:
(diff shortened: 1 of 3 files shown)


NetDwarf:

Storm breaks with this PR. As far as I can tell it's probably a necessary change in Storm, but I am currently a bit too ill to think straight. Just a small heads-up. :slightly_smiling_face:

tegstewart:
Storm breaking is weird.  That implies there are NPCs with the protect ability, and whose brain applies it.  Mercenary/henchmen mobs, maybe?

NetDwarf:
> Storm breaking is weird. That implies there are NPCs with the protect ability, and whose brain applies it. Mercenary/henchmen mobs, maybe?

Have you played the custom dungeons in Storm? I think it is the Midwinter dungeon with the wonky aggro (among others). That's probably why, but I might be totally wrong.

tegstewart:
I haven't done much on Storm, and was thinking a few days ago I should spend some time on it.

If anything, previous code should have been more likely to break things.  The effect was being throwing an exception that was being caught when the effect was applied, but throwing an unhandled exception when the effect was removed.

PR #489, opened 2024-11-05
Fix AddToAggroList() LOS Checks

tegstewart:
I noticed PvE keep lords calling BringFriends() and bringing all the keep guards to them before a hostile player was anywhere near them, because keep guards are hardcoded twice over to never check LOS in AddToAggroList().  That appears to cause other problems down the line with keep guard specific code needed to stop them from spamming LOS checks in StartAttack() when they try to attack players who shouldn't be in their aggro tables in the first place.  I've tried to remedy that by adding a server property to enable keep guard LOS in AddToAggroList, which is disabled by default to maintain existing functionality.

In exploring that, I discovered that StandardMobBrain.AddAggroToList()'s LOS check is flawed.  The LOS request is sent to the client, a callback is set up, but the result for all LOS checks are stored in a single local property so only the most recent LOS result is stored, and that property is checked before the callbacks have had time to run.  Each Think() cycle calls CheckNPCAggro() and CheckPlayerAggro(), which make decisions based on the previous cycle's most recent LOS check result, which is then applied to all NPCs and players within aggrorange whether they're in LOS or not.  It's possible the previous LOS check is from a target that is no longer in LOS or even in aggro range.  I expect there are a lot of weird intermittent initial aggro issues stemming from this.

So I've split up AddAggroToList() in a manner akin to StartAttack() and ContinueStartAttack().  If no LOS check is done, AddAggroToList() calls ContinueAddAggroToList(), and behaviour shouldn't change.  If a LOS check is done and there is LOS, CheckAggroLOS() will call ContinueAddAggroToList() and AttackMostWanted() to get the mob to attack immediately instead of waiting for the next Think() cycle.  CheckAggroLOS() doesn't get the aggro amount passed to it, so a list is used to store the pending aggro amounts based on ObjectID.  I've tried to streamline that as much as possible to prevent a race condition in which AddToAggroList() is adding items to the list faster than CheckAggroLOS() is removing them.

There are more issues with keep guards that I need to look into, but I want to make sure the StandardMobBrain LOS checks are correct before I get too far into that.

I expect we're going to want to change a few things before merging this.  I don't have a lot of experience with callbacks and thread safety, so there may be potential issues I've overlooked, or a better way to get aggro amounts into CheckAggroLOS().

### DIFF
--- a/GameServer/ai/brain/BD Pets/BDHealerBrain.cs
+++ b/GameServer/ai/brain/BD Pets/BDHealerBrain.cs
@@ -37,11 +37,6 @@ namespace DOL.AI.Brain
 	public class BDHealerBrain : BDPetBrain
 	{
 		/// <summary>
-		/// Defines a logger for this class.
-		/// </summary>
-		private static readonly ILog log = LogManager.GetLogger(MethodBase.GetCurrentMethod().DeclaringType);
-
-		/// <summary>
 		/// Constructs new controlled npc brain
 		/// </summary>
 		/// <param name="owner"></param>
@@ -214,7 +209,7 @@ namespace DOL.AI.Brain
 		/// </summary>
 		/// <param name="living"></param>
 		/// <param name="aggroamount"></param>
-		public override void AddToAggroList(GameLiving living, int aggroamount) { }
+		public override void AddToAggroList(GameLiving living, int aggroamount, bool CheckLOS = false) { }
 
 		public override void  RemoveFromAggroList(GameLiving living) { }
 
--- a/GameServer/ai/brain/ControlledNpcBrain.cs
+++ b/GameServer/ai/brain/ControlledNpcBrain.cs
@@ -935,7 +935,7 @@ namespace DOL.AI.Brain
 		/// </summary>
 		/// <param name="living"></param>
 		/// <param name="aggroamount"></param>
-		public override void AddToAggroList(GameLiving living, int aggroamount)
+		public override void AddToAggroList(GameLiving living, int aggroamount, bool CheckLOS = false)
 		{
             GameNPC npc_owner = GetNPCOwner();
             if (npc_owner == null || !(npc_owner.Brain is StandardMobBrain))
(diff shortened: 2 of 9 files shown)


NetDwarf:
This conflicts with #488 :sweat_smile: 

This PR is going to need a bit of testing as LOS checking and aggro is finicky in general. How did you test correct behavior? :slightly_smiling_face:

tegstewart:
I use a LOT of log entries to track behaviour in real time, and filter which NPCs I'm tracking on via OID.  Keeps and DF are particularly good in terms of LOS checks, but I suspect parts of Tuscaren Glacier would also work pretty well.

One of the things I noticed was GameKeepGuard.Think() runs CheckPlayerAggro() and CheckNPCAggro(), and then keep guard brains then run it a second time.

I think I'm going to break this into smaller pieces, so it's easier to test each step independently.

I've been working on something else recently, without this change added, and now that I know to look for it, I'm seeing those weird intermittent aggro issues every now and then when using a pet.  If only one of us is in LOS, the mob will  sometimes fail to aggro for a few cycles, and at other times will get aggro to the one out of LOS, turn to face them, but then stall out because the attack LOS fails.

NetDwarf:
> I use a LOT of log entries to track behaviour in real time, and filter which NPCs I'm tracking on via OID. Keeps and DF are particularly good in terms of LOS checks, but I suspect parts of Tuscaren Glacier would also work pretty well.

Would it be possible to add something to the admin/GM tooling via a command or such to debug this kind of issue?

> I've been working on something else recently, without this change added, and now that I know to look for it, I'm seeing those weird intermittent aggro issues every now and then when using a pet. If only one of us is in LOS, the mob will sometimes fail to aggro for a few cycles, and at other times will get aggro to the one out of LOS, turn to face them, but then stall out because the attack LOS fails.

Are the problems you describe with or without this PR?

> One of the things I noticed was GameKeepGuard.Think() runs CheckPlayerAggro() and CheckNPCAggro(), and then keep guard brains then run it a second time.

Yeah, the AI code is really tangled. It's certainly in need of cleaning/refactoring. :sweat_smile:

NetDwarf:
Given the PR cannot be merged I am closing it for now.

PR #490, opened 2024-11-07
Apply Minimum Spell Hit Chance

tegstewart:
We're correctly reducing spell damage when spell hit chance is below the 55% minimum, but were then incorrectly using the <55% hit chance.

Only having a 40% chance to hit Gjal with a spell and then also having damage reduced by 64.5% on top of that is just mean.

### DIFF
--- a/GameServer/spells/SpellHandler.cs
+++ b/GameServer/spells/SpellHandler.cs
@@ -2495,13 +2495,15 @@ namespace DOL.GS.Spells
 			return StartSpell(target);
 		}
 
+        protected const int minHitChance = 55;
+        protected const int maxResistChance = 100 - minHitChance;
 
-		/// <summary>
-		/// Called when spell effect has to be started and applied to targets
-		/// This is typically called after calling CheckBeginCast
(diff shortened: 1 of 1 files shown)


tegstewart:
Bah, Tolakram's notes seem to indicate that 55% claim is flat out not true.  Blech.

I know other private servers went with the 55% cap, so I'm now wondering if this should be a server property, disabled by default.

NetDwarf:
Would that mean that a spell would resist with a chance of `maxResistChance` at most for any mob (even a dragon) as the name suggests?

The double disadvantage against high level mobs is intended. I can say that it is at least not totally un-livelike for the most common cases. It might even be too weak for very high mobs as dragons as even a group of players were mostly unable to hit these with spells without pet spam.

tegstewart:
I'll close this, then.  Wish I'd noticed Tolak's notes before PRing, but they're in a different part of the file than I was working on.

It's not really a problem, anyway.  People can reduce or remove the damage penalty via existing server properties anyway.

